PR #56, opened 2022-08-29
Fix the `tool.hdev.project_type` setting in `pyproject.toml`

seanh:
Fix the `project_type` setting in the `[tool.hdev]` section in `pyproject.toml`: `hdev` uses `library` rather than `package` for packages (including both libraries like `h-vialib` and command line tools like `hdev` itself).

Also simplify this by making the `project_type` variable's value come from a `__hdev_project_type` setting in `cookiecutter.json`, which also means that it can be varied on a per-project basis.

### DIFF
--- a/pypackage/cookiecutter.json
+++ b/pypackage/cookiecutter.json
@@ -17,6 +17,7 @@
     "__pypi_url": "https://pypi.org/project/{{ cookiecutter.slug }}",
     "_extensions": ["local_extensions.LocalJinja2Extension"],
     "_directory": "pypackage",
+    "__hdev_project_type": "library",
     "__ignore__": "",
     "__target_dir__": ""
 }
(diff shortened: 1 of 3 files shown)


pypackage/cookiecutter.json
```diff
@@ -17,6 +17,7 @@
     "__pypi_url": "https://pypi.org/project/{{ cookiecutter.slug }}",
     "_extensions": ["local_extensions.LocalJinja2Extension"],
     "_directory": "pypackage",
+    "__hdev_project_type": "library",
```
seanh, line 20: Existing projects that have already had one of the new cookiecutter applied won't have a `__hdev_project_type` in their `cookiecutter.json`. When that happens the cookiecutter just defaults to the setting from this `cookiecutter.json` file in the cookiecutter itself. But any project can always add an `__hdev_project_type` to its `cookiecutter.json` to override the default

marcospri (approved):

PR #57, opened 2022-08-31
Support Python 3.7 in packages by default

seanh:
Add 3.7 to the list of Python versions supported by default when you create a new package using the pypackage cookiecutter. This means they support all four currently supported versions of Python by default.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.8'
+      - name: Install Python 3.7
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.7'
       - run: python -m pip install cookiecutter tox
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -20,6 +20,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.8'
+      - name: Install Python 3.7
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.7'
```
seanh, line 26: The cookiecutters repo's own `ci.yml` needs to install all versions of Python listed in `pyramid-app/cookiecutter.json` and `pypackage/cookiecutter.json`. We just have to manage this manually, I don't think there's any way to automate this that would be worth the cost.

marcospri (approved):
Oh no, that's going to have a few follow up PRs. Approved :white_check_mark:

PR #71, opened 2022-10-03
Update Python patch versions

seanh:
- Upgrade the Python patch versions used for newly-generated projects
- Fix `make test` on macOS


seanh:
[Approved in Slack](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1664806252832549?thread_ts=1664800379.899079&cid=C4K6M7P5E)

PR #77, opened 2022-11-07
Use the default Python version in the tests

seanh:
If the tests don't specify a Python version then the default one from each cookiecutter's `cookiecutter.json` file will be used. This is good because it means the tests are testing that new projects created using the cookiecutters will work with the Python version(s) that the cookiecutters use by default.


seanh:
Closing in favour of https://github.com/hypothesis/cookiecutters/pull/81

PR #79, opened 2022-11-07
Move `README.md` into `_shared`

seanh:
Replace the two separate `pypackage` and `pyramid-app` `README.md` templates with a single shared template.

This does result in a fairly long `README.md` template with a few `pypackage`-only and `pyramid-app`-only sections. But I think it's worth it to keep the `README.md`'s consistent across different types of project and to make the template easier to maintain, especially since I'm [planning to add a third cookiecutter](https://github.com/hypothesis/cookiecutters/pull/74/files) that will share the same `README.md` template.

### Testing

This shouldn't make any difference to the generated `README.md` for all the existing projects that're already using the `pypackage` cookiecutter. For example using [cookiecutter-pypackage-test](https://github.com/hypothesis/cookiecutter-pypackage-test) as a test:

```console
$ git clone https://github.com/hypothesis/cookiecutter-pypackage-test.git /tmp/cookiecutter-pypackage-test
$ cd /tmp/cookiecutter-pypackage-test/
$ make template checkout=move-README.md-into-shared
$ git diff
$
```

This *does* make a small change to the generated `README.md`'s for projects using the `pyramid-app` cookiecutter because there were some differences in how the `pypackage` and `pyramid-app` `README.md`'s documented the same prerequisites. Fortunately we don't have any projects using that cookiecutter yet other than [cookiecutter-pyramid-app-test](https://github.com/hypothesis/cookiecutter-pyramid-app-test). Also the new `README.md` template only puts `make services` and `make devdata` in the `README.md` if the project _has_ services and devdata and not if it only has do-nothing `make services` and `make devdata` commands. cookiecutter-pyramid-app-test doesn't have services or devdata so they get removed from its `README.md`: 

```console
$ git clone https://github.com/hypothesis/cookiecutter-pyramid-app-test.git /tmp/cookiecutter-pyramid-app-test
$ cd /tmp/cookiecutter-pyramid-app-test/
$ make template checkout=move-README.md-into-shared
$ git diff
diff --git a/README.md b/README.md
index 2f00562..85524c9 100644
--- a/README.md
+++ b/README.md
@@ -13,23 +13,20 @@ First you'll need to install:
 
 * [Git](https://git-scm.com/).
   On Ubuntu: `sudo apt install git`, on macOS: `brew install git`.
+* [GNU Make](https://www.gnu.org/software/make/).
+  This is probably already installed, run `make --version` to check.
 * [pyenv](https://github.com/pyenv/pyenv).
-  See [pyenv's README](https://github.com/pyenv/pyenv#readme) for install instructions.
-  First you need to [install the Python build dependencies](https://github.com/pyenv/pyenv/wiki#suggested-build-environment)
-  then on macOS use the **Homebrew** installation method,
-  on Ubuntu use the **Basic GitHub Checkout** method.
+  Follow the instructions in pyenv's README to install it.
+  The **Homebrew** method works best on macOS.
+  The **Basic GitHub Checkout** method works best on Ubuntu.
   You _don't_ need to set up pyenv's shell integration ("shims"), you can
   [use pyenv without shims](https://github.com/pyenv/pyenv#using-pyenv-without-shims).
-* [GNU Make](https://www.gnu.org/software/make/).
-  This is probably already installed or will have been installed while installing pyenv, run `make --version` to check.
 
 Then to set up your development environment:
 
 ```terminal
 git clone https://github.com/hypothesis/pyramid-app-cookiecutter-test.git
 cd pyramid_app_cookiecutter_test
-make services
-make devdata
 make help
 ```
 
```


_shared/project/README.md
```diff
@@ -0,0 +1,263 @@
+{% if cookiecutter.get("visibility") == "public" -%}
+<a href="{{ cookiecutter.__github_url }}/actions/workflows/ci.yml?query=branch%3Amain"><img src="https://img.shields.io/github/workflow/status/{{ cookiecutter.github_owner }}/{{ cookiecutter.slug }}/CI/main"></a>
+{% endif -%}
+{% if cookiecutter._directory == 'pypackage' -%}
+<a href="{{ cookiecutter.__pypi_url }}"><img src="https://img.shields.io/pypi/v/{{ cookiecutter.slug }}"></a>
+{% endif -%}
```
jon-betts, line 6:
```suggestion
{% if cookiecutter.get("visibility") == "public" -%}
    <a href="{{ cookiecutter.__github_url }}/actions/workflows/ci.yml?query=branch%3Amain"><img src="https://img.shields.io/github/workflow/status/{{ cookiecutter.github_owner }}/{{ cookiecutter.slug }}/CI/main"></a>
{% endif -%}
{% if cookiecutter._directory == 'pypackage' -%}
    <a href="{{ cookiecutter.__pypi_url }}"><img src="https://img.shields.io/pypi/v/{{ cookiecutter.slug }}"></a>
{% endif -%}
```

It might result in slightly less eye murder if we can indent blocks where we can. I think we're stripping whitespace here anyway, so it shouldn't matter?

_shared/project/README.md
```diff
@@ -0,0 +1,263 @@
...
+   repositories.
+
+2. Now that the {{ cookiecutter.slug }} project has access to the `PYPI_TOKEN` secret
+   you can release a new version by just [creating a new GitHub release](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository).
+   Publishing a new GitHub release will automatically trigger
+   [a GitHub Actions workflow](.github/workflows/pypi.yml)
+   that will build the new version of your Python package and upload it to
+   <{{ cookiecutter.__pypi_url }}>.
+{% endif -%}
+
```
jon-betts, line 120: These blocks are so long it's quite hard to see what the ifs and else's correlate to.

I think we might be better off with a separate include blocks here which have the parts for the various projects in separate files. Then you could have something like:

```jinja2
{% if cookiecutter._directory == 'pypackage' %}
    {{ include("hacking/pypackage/release") }}
    {{ include("hacking/pypackage/python_versions") }}
{% elif cookiecutter._directory == 'pyramid-app' %}
    {{ include("hacking/pyramid_app/python_versions") }}
{% endif %}
```

_shared/project/README.md
```diff
@@ -0,0 +1,263 @@
...
+  You **do** need to set up the `docker` command to work without `sudo`,
+  on Linux this means following Docker's [Post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/).
+{%- endif %}
+{%- if cookiecutter.get("frontend") == "yes" %}
+* [Node](https://nodejs.org/) and npm.
+  On Ubuntu: `sudo snap install --classic node`.
+  On macOS: `brew install node`.
+* [Yarn](https://yarnpkg.com/): `sudo npm install -g yarn`.
+{%- endif %}
+{{- include("hacking/prerequisites") }}
```
jon-betts, line 88: Is this `hacking` any more?

_shared/project/README.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+## Uninstalling
+
+To uninstall run:
+
+```
+pipx uninstall {{ cookiecutter.slug }}
+```
+{%- endif %}
```
jon-betts, line 57: Without indentation it's not easy to see what this is an if around? There's a suggestion below about splitting things out into separate files and including the whole lot. You could do something similar here.

_shared/project/README.md
jon-betts: Another thought is if we have this much which is custom to one type of project or another, but is not "shared" in a meaningful way, is this the right place for it?

We could replace this with `{% block custom_content %}` and use `extend` to keep the parts which are specific to pypackages close to the templates they serve.

jon-betts (commented):
I find this a bit confusing to look at, and wouldn't want to have to maintain this. I think this is for a few reasons:

 * Long blocks without indentation are hard to visually identify where they begin and end
 * Long blocks in general are tricky as they can push of the relevant if statement which tells you what they apply to (so you see and "else", but it's hard to know what's on both sides without scrolling).
 * The blend of specific / re-used content seems off to me. We have a lot project type specific documentation inline here

I think we can solve some of this with a combination of techniques:

 * Put long sections in separate files and include them, keeping blocks short
 * Try and indent where we can
 * Think about keeping non-shared components out of the shared template and using `extends`

If we extend this method to a third or fourth project type, I think it could get quite confusing.

seanh:
> * Try and indent where we can

I've added indentation to the one-liner `{% if %}` blocks in this PR because as you pointed out for one-liners Jinja2's whitespace control will remove the indentation. Unfortunately this can't be done for multi-line blocks: whitespace control will only dedent the first line not the rest. I think this is a limitation of Jinja2: IIRC there's no way to indent a block in a Jinja2 template and not have those lines be indented in the output file.

## We can't use `{% include %}`, `{% extends %}`, `{% block %}`, etc

> I think we can solve some of this with a combination of techniques:
> 
> * Put long sections in separate files and include them, keeping blocks short
> * Think about keeping non-shared components out of the shared template and using `extends`

These unfortunately aren't possible. Due to the limitations of cookiecutter one template file within the `cookiecutters` repo can't really include or extend another. `{% extends %}`, `{% include %}` and `{% import %}` all depend on the `searchpath` argument to Jinja2's [`FileSystemLoader`](https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.FileSystemLoader): you have to pass it the path or list of paths where it should look for the template files that're being extended or included. And unfortunately [cookiecutter is hard-coded to pass no `searchpath` to `FileSystemLoader`](https://github.com/cookiecutter/cookiecutter/blob/f9376a96097086476ce9eb0b93297a471ae520e0/cookiecutter/generate.py#L318) and this isn't configurable.

So `{% extends %}`, `{% include %}` and `{% import %}` are disabled in cookiecutter. `{% block %}`, `{% super %}` and other related features can't be used either because they depend on `{% extends %}`.

I looked into this when I first created this cookiecutters repo, I couldn't see a way to get it working. I spoke to the cookiecutter and Jinja2 developers at the time and they confirmed that there isn't a way.

<details>
<summary>

## There's a horrible hack to get them working put I don't think we should do it
</summary>

Actually there is a horrible hack to use `{% include %}` or `{% extends %}` but it's weird enough that I think we're better off not using it:

* Cookiecutter actually does pass one search path to `FileSystemLoader`: `"."` which is the template directory of the particularly cookiecutter being used, `pypackage/{{ cookiecutter.slug }}/`, `pyapp/{{ cookiecutter.slug }}/` or `pyramid-app/{{ cookiecutter.slug }}/`
* Any files in those directories get rendered into the project. So if you add a `pyapp/{{ cookiecutter.slug }}/includes/foo.md` or a `pyapp/{{ cookiecutter.slug }}/base/README.md` those include files (`includes/foo.md`) and base files (`base/README.md`) would themselves get rendered directly into the project by cookiecutter.
* We could use [our `paths_to_remove` feature](https://github.com/hypothesis/cookiecutters/blob/8ddef42341dccc701711d4fe31462e034e75acd4/_shared/hooks/post_gen_project.py#L12-L42) to remove those files from the projects.
* It's weird that the include and base files would need to live in each of the individual cookiecutters, not in the `_shared/` directory. If we want to split `_shared/project/README.md` into multiple files for example then we could put includes like `{% include "includes/README/installing.md" %}` but even though this `installing.md` file is included by the `README.md` template in the `_shared` directory the `installing.md` file itself would need to live in the individual cookiecutter directories (in triplicate for the three cookiecutters). We could avoid duplicating `installing.md` by putting in in `_shared/includes/installing.md` but then each of the three cookiecutters would need to contain a symlink to each shared include or base file. Cookiecutter would then render the include and base files into the project so we'd have to use `post_gen_project.py` to remove them.

I think the cure would be worse than the disease. We could move `README.md`'s long `{% if %}`'s into separate files but the way we'd have to do that would be pretty twisted, with files and symlinks forced to live in odd places and post-generation code for removing files from the generated project. The fact that you can get this to work is not intended by cookiecutter, we'd be abusing an unintended implementation side-effect.

I understand that the large `{% if %}`'s aren't very readable. But it's a very simple approach. It's easy to understand how it works: there is one `README.md` template, that template contains some `{% if %}`'s, that's all there is to it. This is much better than crazy hackery involving intimate knowledge of unintended cookiecutter implementation side-effects.
</details>

<details>
<summary>

## There's an unreleased cookiecutter PR about this but I don't think it's useful
</summary>

They actually merged a PR in June to add `"../shared"` to the search paths: https://github.com/cookiecutter/cookiecutter/pull/1485 But they haven't released a new version of cookiecutter since merging that. It's of limited use to us anyway. The include and base files would still need to live in the individual cookiecutters not in the shared directory. It's just that they can now live in `pyapp/templates/`, `pypackage/templates/` and `pyramid-app/templates/`, meaning they're outside of the `{{ cookiecutter.slug }}` directory so cookiecutter won't render the include and base files into the projects and our `post_gen_project.py` won't need to remove them. But you still have the problem that `README.md` is in `_shared` but any files it wants to extend or include must live in each one of the individual cookiecutters. And this PR is unreleased anyway.

</details>

## We could use macros

So I think we're limited to within-a-single-file techniques. I think [macros](https://jinja.palletsprojects.com/en/3.1.x/templates/#macros) may be all that's available to us. I think macros would maintain the simplicity that the current `README.md` has with `{% if %}`'s: it's just one `README.md` template and it calls some macros. All in one (admittedly long) file, simple to understand what's going on, no crazy cookiecutter hacks. Macros are limited in how much they help us however:

* Macros would enable us to define the various parts that are specific to one cookiecutter or another in a sequence of macros at the top of the file, and then have a more concise "map" of the file (including all the `{% if %}` and `{% else %}` logic) at the bottom.
* You'd still have long macros and they still can't be indented, so the "Am I in an `{% if %}?`" problem arguably just becomes "Am I in a macro?". The fact that all the macros are at the top of the file eases this somewhat
* We can't use macros to break up the large file into smaller ones: all the macros would need to be defined within the file since we can't use `{% import %}`
* It arguably makes readability worse in some ways: the output contents would now be out-of-order within the template file since some parts have been moved into macros at the top. We could mitigate this by moving *every* section into a macro, even the unconditional ones
* It's arguably weird usage to define a bunch of macros that are each only used once and take no arguments. But it lets us split the file between contents (the macros) and map

Overall my feeling is that breaking up `README.md` into macros might be an improvement, worth a try...

_shared/project/README.md
seanh: This happened when I renamed `HACKING.md` to `README.md`: I didn't want to rename the include file as then I'd have to rename that file in every project that uses it. We can do that but I think it's a separate PR (this include is already on `main`, this PR just moves it into the shared dir).

_shared/project/README.md
seanh: Responded to this in the [long comment below](https://github.com/hypothesis/cookiecutters/pull/79#issuecomment-1307220125). TLDR: I don't think we can split the `README.md` template into separate files: we can use `include`, `extend` or `block`. But I think breaking it up into macros still within a single file might help with this.

I think macros would allow us to have a concise "map" of the logic of the file similar to the example you posted:

```jinja2
{% if cookiecutter._directory == 'pypackage' %}
    {{ release() }}
    {{ package)python_versions() }}
{% elif cookiecutter._directory == 'pyramid-app' %}
    {{ app_python_versions() }}
{% endif %}
```

It's just that all the macros need to be defined at the top of the `README.md` file (they can't be defined in other files) and then the logical map would be at the bottom of the file. See the longer comment below for details.

_shared/project/README.md
seanh: Sure. I've done this for the one-liners but I don't think this is possible for multi-line blocks: whitespace control only dedents the first line not the rest. IIRC this is a limitation of Jinja2: there's no way to indent a multiline block in the template without it getting indented in the output file

_shared/project/README.md
seanh: See https://github.com/hypothesis/cookiecutters/pull/79#issuecomment-1307238767 for the mix of shared and specific content

seanh:
I've spent a couple of hours playing with macros and indentation and I'm gonna say the the original `{% if %}`-based approach is by far the lesser of two evils here:

* Macros by themselves (without indentation) don't make the template any more readable. You just have all the same long blocks but in macros and now with additional code at the end to actually call the macros.

  Also it really plays hell with whitespace control: the section at the top of the file containing the actual macros introduces a load of blank lines into the output file (the blank lines between the macros and the newlines before and after the `{% macro ... %}` and `{% endmacro %}` lines). If you want any blank lines between all the `{% if %}`'s and `{% else %}`'s and macro calls at the bottom of the file to make that Jinja code more readable then those blank lines will get into the output file as well (and so will the new lines before and after the `{% if %}`, `{% else %}` and `{% endif %}` lines). It *is* possible to get the output right but it's very fiddly and the result is much harder to maintain than the `{% if %}`'s.

* If you then indent the contents of the macros it's even worse. Jinja2 isn't meant to do this: it doesn't have a `dedent` filter or allow you to indent the contents of a block without that text getting indented in the output file. If you try to add it you get into a fight with Jinja2. Jinja2's whitespace control will dedent the first line of a macro which will then mean that Python's `dedent()` (called by your custom `dedent` filter that you've added) will not dedent the rest of the lines (since there's no longer a common indentation shared by all lines including the first one). But if you remove the Jinja2 whitespace control then that fixes the dedenting but also introduces unwanted blank lines into the file.

Basically the whitespace control and dedentation needed to get this working is extremely fiddly and difficult to understand, and that's a much worse maintainability problem than mere long `{% if %}` blocks.

My suggestion at this point: merge this PR as-is with the `{% if %}`'s and consider in the future splitting up `README.md` into a few separate `docs/*.md` files, probably something like.

seanh:
Actually scratch all that, I found a couple of non-default settings that make Jinja2 behave much more sanely around whitespace: https://github.com/hypothesis/cookiecutters/pull/82. With that done it's now fairly easy to break `README.md` into indented macros without introducing a lot of unwanted whitespace changes. I pushed a couple of commits to this PR to do that. But this now depends on https://github.com/hypothesis/cookiecutters/pull/82

_shared/project/README.md
```diff
@@ -0,0 +1,294 @@
...
+{% endif %}
+{% if cookiecutter._directory == 'pyramid-app' -%}
+    {{ app_python_dependencies()|dedent }}
+{%- else -%}
+    {{ package_python_dependencies()|dedent }}
+{%- endif %}
+{% if include_exists("README/tail.md") %}
+
+{{ include("README/tail.md") -}}
+{% endif %}
```
jon-betts, line 294:
```suggestion
{{ badges() | dedent }}
{{ intro() | dedent }}
{% if cookiecutter.get("console_script") == "yes" -%}
    {{ installing() | dedent }}
{% endif %}
{{ setting_up() | dedent }}
{% if cookiecutter._directory == 'pypackage' -%}
    {{ releasing() | dedent }}
{% endif %}
{% if cookiecutter._directory == 'pyramid-app' -%}
    {{ app_python_version() | dedent }}
{% else -%}
    {{ package_python_version() | dedent }}
{% endif %}
{% if cookiecutter._directory == 'pyramid-app' -%}
    {{ app_python_dependencies() | dedent }}
{%- else -%}
    {{ package_python_dependencies() | dedent }}
{%- endif %}
{% if include_exists("README/tail.md") %}
    {{ include("README/tail.md") -}}
{% endif %}
```

Mostly putting some space around the pipe and indenting the final part.

jon-betts (approved):
I think this is a decent compromise here. The long blocks at the top of the read me are what they are, but the result is there is somewhere you can go right at the bottom to see the layout of the document and re-arrange it if you wanted to.

seanh:
Rebased

_shared/project/README.md
```diff
@@ -0,0 +1,294 @@
+{%- macro badges() %}
+    {% if cookiecutter.get("visibility") == "public" %}
+    <a href="{{ cookiecutter.__github_url }}/actions/workflows/ci.yml?query=branch%3Amain"><img src="https://img.shields.io/github/workflow/status/{{ cookiecutter.github_owner }}/{{ cookiecutter.slug }}/CI/main"></a>
+    {% endif %}
+    {% if cookiecutter._directory == 'pypackage' %}
+    <a href="{{ cookiecutter.__pypi_url }}"><img src="https://img.shields.io/pypi/v/{{ cookiecutter.slug }}"></a>
+    {% endif %}
```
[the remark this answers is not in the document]
seanh: Discussed elsewhere but unfortunately no, you can't indent these lines in the template without them becoming indented in the output file. This is just how Jinja2 works

_shared/project/README.md
seanh: You can't indent that final `{{ include("README/tail.md") -}}` because there's more than one line in the body of the `{% if %}`: there's a blank line in the `{% if %}`-body above the `{{ include("README/tail.md") -}}`, the blank line is needed because we do want an empty line to appear in the output file before the contents of the `tail.md` file. Discussed elsewhere but generally speaking you can't indent lines in Jinja2 templates if you don't want those lines to be indented in the output. The custom `| dedent` filter here allows us to get away with indenting macro bodies but doesn't apply to blocks in general like `{% if %}` bodies.

I'm gonna leave the `|`'s without spaces around them as that's the style used in the Jinja2 docs

_shared/project/README.md
```diff
@@ -0,0 +1,294 @@
...
+{% endmacro %}
+
+{%- macro intro() %}
+    # {{ cookiecutter.name }}
+
+    {{ cookiecutter.short_description }}
+{% if include_exists("README/head.md") %}
+
+{{ include("README/head.md", indent=4) -}}
+{% endif %}
```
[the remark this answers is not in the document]
seanh: I've moved this out into the logic block at the bottom, makes more sense there

PR #82, opened 2022-11-08
Use Jinja2's `trim_blocks` and `lstrip_blocks`

seanh:
This makes Jinja2 automatically remove whitespace around template tags meaning much less manual whitespace management is needed. When using `{% if %}`'s etc it generally just does what we want 99% of the time and 1% of the time we have to use explicit whitespace control, rather than the other way around.

I also made a change to our custom `include()` function. Previously you used to be able to just do this:

```jinja2
{{- include("foo/bar") }}
```

and that would render the contents of the project's `.cookiecutter/includes/foo/bar` file if it exists or would render nothing (not even a blank line) if the include file doesn't exist.

This is not normal Jinja2 behaviour. To get this to work the `include()` Python code would prepend a `"\n"` to the text from the include file and would also strip any trailing whitespace (including the `"\n"` at the end of the last line of the file). So it'd remove the `"\n"` from the end of the file and prepend a `"\n"` to the beginning!

It didn't actually work either. In many cases it would remove any blank lines that were above the `include()` in the template. So if you wanted to have a blank line before the contents of an include file the template couldn't do that. You had to actually put a blank line at the top of the include file in every project.

This hack no longer works at all with `trim_blocks` and `lstrip_blocks`: the output gets messed up any time you try to use an include.

Also having `include()` mess with newlines like this is a really bad idea because it's so unexpected and confusing. I forgot that I had done it, and it took me ages wondering why the flip whitespace control was going nuts before I finally realised that our custom Python code was doing this.

I've simplified `include()` to not do anything unexpected: it neither prepends a leading `"\n"` nor strips a trailing one.

This does make the templates a bit more wordy as they frequently have to do variations on this pattern:

```jinja2
{% if include_exists("foo/bar") %}
    {{ include("foo/bar") }}
{% endif %}
```

But at least this is comprehensible. It lets the template freely use Jinja2's whitespace control go get whatever whitespace behaviour you want (`-` or no `-`, leading or trailing blank lines in the template or not, etc). It gives the same whitespace behaviour as you'd get with other conditional blocks like:

```jinja2
{% if cookiecutter.get("services") %}
.PHONY: services
services: args?=up -d
services: python
	@tox -qe dockercompose -- $(args)
{% endif %}
```

I also changed `include()` to use the stdlib's `indent()` instead of implementing its own indent code.

### DIFF
--- a/_shared/local_extensions.py
+++ b/_shared/local_extensions.py
@@ -1,3 +1,4 @@
+import textwrap
 from pathlib import Path
 
 from jinja2.ext import Extension, pass_context
@@ -109,9 +110,9 @@ class LocalJinja2Extension(Extension):
         try:
             lines = self._readlines(context, path)
         except (FileNotFoundError, NotADirectoryError):
-            lines = []
+            return ""
 
-        return "".join(["\n"] + [(" " * indent) + line for line in lines]).rstrip()
+        return textwrap.indent("".join(lines), " " * indent)
 
     def oldest(self, python_versions):
         """Return the oldest from `python_versions` by version number."""
--- a/_shared/project/.github/workflows/ci.yml
+++ b/_shared/project/.github/workflows/ci.yml
@@ -2,18 +2,18 @@ name: CI
 on:
   push:
   workflow_dispatch:
-  {%- if cookiecutter._directory == 'pypackage' %}
+  {% if cookiecutter._directory == 'pypackage' %}
   schedule:
   - cron: '0 1 * * *'
-  {%- endif %}
+  {% endif %}
 jobs:
-  {%- for job in ['format', 'lint', 'tests', 'coverage', 'functests'] %}
+  {% for job in ['format', 'lint', 'tests', 'coverage', 'functests'] %}
   {{ job|capitalize }}:
-    {%- if job == 'coverage' %}
+    {% if job == 'coverage' %}
     needs: tests
-    {%- endif %}
+    {% endif %}
     runs-on: ubuntu-latest
-    {%- if cookiecutter.get("db") == "yes" %}
+    {% if cookiecutter.get("db") == "yes" %}
     services:
       postgres:
         image: postgres:11.5-alpine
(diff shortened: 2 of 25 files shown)


seanh:
I'll need to go through and test all the various options and includes to make sure all the formatting is just right in every case. But this seems to be a big improvement

seanh:
I think this is ready to go. I've tested it pretty thoroughly making sure the output is right when each option is on or off and when each include is used or not used, and testing that it won't make any unwanted changes to existing projects.

_shared/local_extensions.py
```diff
@@ -109,9 +110,10 @@ def include(self, context, path, indent=0):
         try:
             lines = self._readlines(context, path)
         except (FileNotFoundError, NotADirectoryError):
-            lines = []
+            return ""
 
-        return "".join(["\n"] + [(" " * indent) + line for line in lines]).rstrip()
+        s = "".join(lines)
+        return textwrap.indent(s, " " * indent)
```
jon-betts, line 116:
```suggestion
        return textwrap.indent("".join(lines), " " * indent)
```

_shared/project/.github/workflows/ci.yml
```diff
@@ -99,5 +98,5 @@ jobs:
       run: make frontend-lint
     - name: Test
       run: make frontend-test
-  {%- endif %}
-  {{- include(".github/workflows/ci.yml", indent=2) }}
+  {% endif %}
+  {{- include(".github/workflows/ci.yml", indent=2) -}}
```
jon-betts, line 102: I'm surprised this doesn't cause trouble in YAML files, but it looks like we were fighting a lot of whitespace stuff here anyway

_shared/project/setup.cfg
```diff
@@ -20,25 +20,26 @@ packages = find:
 python_requires = >={{ python_versions()|oldest|pyformat(PyFormats.MAJOR_DOT_MINOR_FMT) }}
 install_requires =
     importlib_metadata;python_version<"3.8."
-    {%- if include_exists("setuptools/install_requires") %}
-      {{- include("setuptools/install_requires", indent=4) }}
-    {%- endif %}
+{% if include_exists("setuptools/install_requires") %}
+{{ include("setuptools/install_requires", indent=4) -}}
+{% endif %}
```
jon-betts, line 25:
```suggestion
{% if include_exists("setuptools/install_requires") %}
    {{ include("setuptools/install_requires", indent=4) -}}
{% endif %}
```

Can we indent now we have whitespace stripping?

_shared/project/setup.cfg
```diff
@@ -20,25 +20,26 @@ packages = find:
...
 where = src
 
 [options.entry_points]
-{%- if include_exists("setuptools/console_scripts") %}
+{% if include_exists("setuptools/console_scripts") %}
 console_scripts =
-    {{- include("setuptools/console_scripts", indent=4) }}
-{%- elif cookiecutter.get("console_script") == "yes" %}
+{{ include("setuptools/console_scripts", indent=4) -}}
+{% elif cookiecutter.get("console_script") == "yes" %}
```
jon-betts, line 34: Similar question about indentation here etc.

_shared/project/tox.ini
```diff
@@ -55,26 +59,31 @@ deps =
     lint,tests,functests: factory-boy
     lint,tests,functests: h-matchers
     lint,template: cookiecutter
-{%- endif %}
-    {{- include("tox/deps", indent=4) }}
+{% endif %}
+    {% if include_exists("tox/deps") %}
+        {{- include("tox/deps", indent=4) -}}
+    {% endif %}
```
jon-betts, line 65:
```suggestion
    {% if include_exists("tox/deps") %}
        {{ include("tox/deps", indent=4) }}
    {% endif %}
```

Do we need the whitespace stripping bits here?

jon-betts (commented):
I have a couple of questions about this.

 * There are quite a few places where we have if blocks with statements inside which now have white space control added to them `{{-` I would think you wouldn't need the minuses any more?
 * There are quite a few places where we now don't indent blocks we used to. I thought a motivating factor here was to enable indentation?

This might be an idea for another PR, but I currently find it hard to know what's customizable in a template, or how you would configure it. I think this is because we've chosen to:

 * Have a missing file indicate no customization
 * Have no indicators in the file that there is a customization point

This means a file which has no customization possible looks identical to one which has customization which happens to be un-used. To tackle this I would probably prefer both the customization files to always be present in a project, even if they are empty, and to have some kind of comment or indication. I realise the second part might be tricky as the comment style would need to be configurable.

However, if the files were always present, rather than the presence indicating customzation this would also remove the need to do `include_exists` to check for them.

If the files were always present it would also give a simple solution to adding a comment in the correct style: Each file could start with a comment describing what you should customize in there.

I think this is probably essential to the long term health of our projects otherwise people will edit templated files simply because they don't know any better. I've already done this a bunch of times, as without total knowledge of all the config points there's nothing there to help you know they exist.

_shared/project/tox.ini
seanh: Unfortunately we do need them. This:

```jinja2
Before
{% if include_exists("foo") %}
    {{ include("foo", indent=4) }}
{% endif %}
After
```

outputs this:

```
Before
        included_line_1
    included_line_2
    included_line_3

After
```

The leading `{{-` on the `include()` expression is needed to dedent the three lines from the `tox/deps` file. Alternatively we could not indent the `{{ include("tox/deps", indent=4) }}` line within the `{% if %}`. The trailing `-}}` is to get rid of the unwanted blank line after the `tox/deps` contents.

This is normal Jinja2 behaviour. For example if you were rendering a macro instead of calling our `include()` function it'd be the same:

```jinja2
{% macro foo() %}
macro_line_1
macro_line_2
macro_line_3
{% endmacro -%}

Before
{% if true %}
    {{- foo() -}}
{% endif %}
After
```

_shared/project/setup.cfg
seanh: Since this is a single-line block I think we should be able to indent this one if we used a `{{-`, yes. I'm not sure which approach I want to go with for consistency: always indent single-line block bodies with a `{{-`? Or never indent them so that you don't need the `-`? Multi-line block bodies cannot be indented.

_shared/project/setup.cfg
seanh: It's not possible to indent this because the body of the `{% if %}` contains more than one line. If you simply indent the `{{ include(...) -}}` line that will indent the first line from the `setuptools/console_scripts` include file. If you indent the `include` line with a `{{-` then it'll run that line into the `console_scripts =` line above.

_shared/local_extensions.py
seanh: :facepalm: :+1:

seanh:
> * There are quite a few places where we have if blocks with statements inside which now have white space control added to them `{{-` I would think you wouldn't need the minuses any more?

This is to dedent intended Jinja2 `{{- expression }}`'s. The `lstrip_blocks` setting strips whitespace from the left of `{% ... %}`'s) but not from `{{ ... }}`'s. For example even with `trim_blocks` and `lstrip_blocks` this:

```jinja2
{% if true %}
    {{ "foo " }}
{% endif %}
```

will output this:

```
    foo
```

Notice that the `{% if %}` and `{% endif $}` have not left blank lines above and below the `foo` line, so that's good at least. But the `foo` is still indented. If you don't want `foo` to be indented then you can either do this:

```jinja2
{% if true %}
{{ "foo " }}
{% endif %}
```

or you can use a `{{-` like this:

```jinja2
{% if true %}
    {{- "foo " }}
{% endif %}
```

But **this `{{-` only works for single-line blocks**. If the body of the `{% if %}` is more than one line (even if it just contains a blank line above or below the `foo` line) then you cannot indent the `{% if %}` body. A `{{- "foo" }` would dedent the `foo` and delete any blank line(s) before the `foo`, or if there were a non-blank line before the `foo` then a `{{-` would run the foo into that line, and/or if there are any line(s) after the `foo` they won't be dedented. So `{{-` happens to work for allowing single-line block bodies to be indented but it's a total fail with multi-line blocks.

To be honest I actually wonder if we should stop using `{{-` to indent single-line block bodies and should just dedent them. Then at least it's consistent: block bodies are never indented in the template, whether single- or multi-line. Indentation in the template also means indentation in the output. This is how Jinja2 is meant to be used.

> * There are quite a few places where we now don't indent blocks we used to. I thought a motivating factor here was to enable indentation?

Sadly not. The motivation for this PR is just to get saner whitespace behaviour by enabling `trim_blocks` and `lstrip_blocks`. It means that 99% of the time plain `{% ... %}`'s and `{{ ... }}`'s do what we want, and we only need to use manual whitespace control (`-`'s or `+`'s) 1% of the time. And also the whitespace behaviour is just much less confusing, much more what you would expect.

This change _did_ enable us to do the macro-based indentation that we did in https://github.com/hypothesis/cookiecutters/pull/79. But that indentation only works because we wrap the line in a macro and then pipe the macro to a custom `dedent` filter:

```jinja2
{% macro foo() %}
    Line 1
    Line 2
    Line 3
{% endmacro -%}

{{ foo()|dedent -}}
```

I wouldn't want to use this macro-based solution everywhere. Partially because I wouldn't want to move the bodies of every `{% if %}`, `{% for %}` etc out into macros. And partially because it's honestly kind of a hack--this isn't how Jinja2 is meant to be used, indentation in the template file is meant to correspond to indentation in the output file.

Unfortunately I think we have to give up on the idea of using indentation to show the logical structure of the template like you would in a programming language. I think people are used to doing that in Jinja2 templates because they're writing templates for HTML. It actually totally messes up the indentation of the output HTML but that doesn't matter because indentation doesn't affect HTML rendering and nobody reads the output HTML. But when templating the kind of files that the cookiecutter templates we're forced to reckon with how Jinja2 is meant to be used: the template code itself is indentation-free and the indentation in the template file is meant to follow the indentation of the output file not the logic of the templates.

I wonder if we might want to adopt a style that some people use where all Jinja2 tags are completely unindented (they literally all begin from column 0) and then the non-tag lines are indented as desired for the output. This matches with what's actually going on: the template code is indentation-free and the indentation of the content lines are as desired in the output file. And you then use indentation _within_ the `{% ... %}`'s to express the logical structure of the template, as this doesn't affect rendering:

```jinja2
{% if true %}
{%   if true %}
True line 1
True line 2
True line 3
{%   else %}
False line 1
False line 2
False line 3
{%   endif %}
{% endif %}
```

> This might be an idea for another PR, but I currently find it hard to know what's customizable in a template, or how you would configure it. I think this is because we've chosen to:
> 
> * Have a missing file indicate no customization
> * Have no indicators in the file that there is a customization point

Definitely a separate discussion. This PR is just enabling `trim_blocks` and `lstrip_blocks`.

jon-betts (approved):

_shared/project/setup.cfg
seanh: I've gone through and indented the expression with a `{{-` in all cases where we have a block whose body is just a single expression line (there's only a handful of opportunities to do this)

PR #86, opened 2022-11-11
Speed up `make coverage` by using tox's parallelism

seanh:
`make coverage` currently runs the unit tests for all Python versions in serial and then runs the coverage report. Speed it up by having it run the unit tests in parallel instead, then the coverage report. This is the same as how `make sure` already works.

**Scope:** this PR just speeds up `make coverage` by making it run tests in parallel rather than serial. It doesn't change what `make coverage` actually does. In particular: `make coverage` always runs all the tests before printing the coverage report. It might be better if it only ran the tests if they need to be run (if the code has changed since the last time the tests for each Python version were run and generated coverage data) but getting it to do that is a more vexxed problem than it may seem and anyway :man_shrugging: Note that if you just want to run the coverage report and not the tests you can do `tox -e coverage`. Anyway, out of scope here.

### DIFF
--- a/_shared/project/Makefile
+++ b/_shared/project/Makefile
@@ -88,7 +88,6 @@ frontend-checkformatting:
 
 .PHONY: test
 $(call help,make test,"run the unit tests in Python {{ python_versions()|first|pyformat(PyFormats.MAJOR_DOT_MINOR_FMT) }}")
-coverage: test
 test: python
 	@pyenv exec tox -qe tests
 
@@ -105,7 +104,6 @@ frontend-test:
 {% set target = 'test-py%s'|format(python_version|pyformat(PyFormats.MAJOR_MINOR_FMT)) %}
 .PHONY: {{ target }}
 $(call help,make {{ target }},"run the unit tests in Python {{ python_version|pyformat(PyFormats.MAJOR_DOT_MINOR_FMT) }}")
-coverage: {{ target }}
 {{ target }}: python
 	@pyenv exec tox -qe py{{ python_version|pyformat(PyFormats.MAJOR_MINOR_FMT) }}-tests
 {%- endfor %}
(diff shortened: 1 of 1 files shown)


_shared/project/Makefile
```diff
@@ -105,15 +104,18 @@ frontend-test:
...
 
 .PHONY: coverage
 $(call help,make coverage,"run the tests and print the coverage report")
 coverage: python
-	@pyenv exec tox -qe coverage
+{%- if cookiecutter._directory == 'pyramid-app' %}
+	@pyenv exec tox -qe 'tests,coverage'
+{%- else %}
+	@pyenv exec tox --parallel -qe 'tests,py{{ "{" }}{{ python_versions()|rest|pyformat(PyFormats.MAJOR_MINOR_FMT)|separator(",") }}{{ "}" }}-tests,coverage'
+{%- endif %}
```
seanh, line 118: tox is aware of the dependency of `coverage` on `{pyXY-}tests`. If you just run `tox -e coverage` it'll just run the coverage report. But if you run `tox -e coverage,tests,py39-tests,py38-tests,py37-tests` it'll reorder the environments for you to run `coverage` last. If you add `--parallel` to that then it knows that all the tests can be run in parallel but `coverage` can't start until the tests finish.

_shared/project/Makefile
```diff
@@ -105,15 +104,18 @@ frontend-test:
...
 {{ target }}: python
 	@pyenv exec tox -qe py{{ python_version|pyformat(PyFormats.MAJOR_MINOR_FMT) }}-tests
 {%- endfor %}
 
 .PHONY: coverage
 $(call help,make coverage,"run the tests and print the coverage report")
 coverage: python
-	@pyenv exec tox -qe coverage
+{%- if cookiecutter._directory == 'pyramid-app' %}
+	@pyenv exec tox -qe 'tests,coverage'
```
seanh, line 115: No point in adding `--parallel` here since we're only running two environments and `coverage` can't start until `tests` finishes

marcospri (approved):

PR #91, opened 2022-11-15
Improve the `supervisord[-dev].conf` templates

seanh:
This enables the project to contain a
`.cookiecutter/includes/supervisor.json` file specifying the Supervisor programs to dev and prod. For example for h-h_periodic:

```json
{
  "programs": {
    "h-beat": {
      "dev": true,
      "prod": true,
        "settings": {
          "command": "celery -b %(ENV_H_BROKER_URL)s -A h_periodic.h_beat beat --pidfile=h-celerybeat.pid --loglevel INFO"
        }
    },
    "checkmate-beat": {
      "dev": true,
      "prod": true,
      "settings": {
        "command": "celery -b %(ENV_CHECKMATE_BROKER_URL)s -A h_periodic.checkmate_beat beat --pidfile=checkmate-celerybeat.pid --loglevel INFO"
        }
    },
    "lms-beat": {
      "dev": true,
      "prod": true,
      "settings": {
        "command": "celery -b %(ENV_LMS_BROKER_URL)s -A h_periodic.lms_beat beat --pidfile=lms-celerybeat.pid --loglevel INFO"
      }
    }
  }
}
```


_shared/project/conf/supervisord-dev.conf
```diff
@@ -1,29 +1,32 @@
...
-stderr_events_enabled=true
-stopsignal = KILL
-stopasgroup = true
+{% if include_exists("supervisor.json") %}
+{% set programs = include_json("supervisor.json").get("programs", {}) %}
+{% else %}
+{% if cookiecutter._directory == 'pyramid-app' %}
+{% set programs = {"web": {"dev": true, "prod": true, "settings": {"command": "gunicorn --bind :{{ cookiecutter.port }} --workers 1 --reload --timeout 0 --paste conf/development.ini"}}} %}
+{% else %}
+{% set programs = {"app": {"dev": true, "prod": true, "settings": {"command": "python3 -m h_periodic"}}} %}
```
seanh, line 11: The default programs for non-web apps that don't have a `supervisor.json` file. Whoops, need to not hard-code `h_periodic` here.

_shared/project/conf/supervisord-dev.conf
```diff
@@ -1,29 +1,32 @@
...
-command=python3 -m {{ cookiecutter.package_name }}
-stdout_events_enabled=true
-stderr_events_enabled=true
-stopsignal = KILL
-stopasgroup = true
+{% if include_exists("supervisor.json") %}
+{% set programs = include_json("supervisor.json").get("programs", {}) %}
+{% else %}
+{% if cookiecutter._directory == 'pyramid-app' %}
+{% set programs = {"web": {"dev": true, "prod": true, "settings": {"command": "gunicorn --bind :{{ cookiecutter.port }} --workers 1 --reload --timeout 0 --paste conf/development.ini"}}} %}
```
seanh, line 9: The default programs for Pyramid apps

_shared/project/conf/supervisord-dev.conf
```diff
@@ -1,29 +1,32 @@
...
+{% set default_program_settings = {"stdout_events_enabled": "true", "stderr_events_enabled": "true", "stopsignal": "KILL", "stopasgroup": "true"} %}
+{% for setting_name in default_program_settings.keys()|sort %}
+{{ setting_name }}={{ program_settings.pop(setting_name, default_program_settings[setting_name]) }}
+{% endfor %}
+{% for setting_name, setting_value in program_settings.items() %}
+{{ setting_name }}={{ setting_value }}
+{% endfor %}
 
+{% endif %}
+{% endfor %}
```
seanh, line 29: Ugh, maybe there's a nicer way to write this template code? But Jinja2 doesn't make it easy

_shared/project/conf/supervisord.conf
```diff
@@ -3,29 +3,32 @@ nodaemon=true
...
 logfile_maxbytes=0
-{% if cookiecutter.get("_directory") == "pyapp" %}
 
-[program:app]
-command=python3 -m {{ cookiecutter.package_name }}
-stdout_logfile=NONE
-stderr_logfile=NONE
-stdout_events_enabled=true
-stderr_events_enabled=true
+{% if include_exists("supervisor.json") %}
```
seanh, line 7: With cookiecutter there's no way to share template code or macros between different files within the `cookiecutters` repo, so we end up having to duplicate this whole chunk of template code

_shared/project/conf/supervisord.conf
```diff
@@ -3,29 +3,14 @@ nodaemon=true
...
-{% endif %}
-{% if include_exists("conf/supervisord.conf") %}
-
-{{ include("conf/supervisord.conf") -}}
-{% endif %}
+{% for program_name, program_data in supervisor_settings().programs.items() %}
+[program:{{ program_name }}]
+{% for setting_name, setting_value in program_data.settings.items() %}
+{{ setting_name }}={{ setting_value }}
+{% endfor %}
```
seanh, line 11: This for-loop has to be duplicated because with cookiecutter there's no way to share template or macro code between files. But at least the Python `supervisor_settings()` method has done most of the work so the duplicated for-loop is trivial.

_shared/local_extensions.py
```diff
@@ -110,11 +113,74 @@ def include_exists(self, context, path):
...
+            assert context.name.endswith("supervisord-dev.conf")
+            test = lambda program_settings: program_settings["dev"]
+
+        settings["programs"] = {
+            program_name: program_settings
+            for program_name, program_settings in settings["programs"].items()
+            if test(program_settings)
+        }
+
+        return settings
```
seanh, line 183: This Supervisor-specific method is implemented in Python partially because it's called by two templates (`supervisord.conf` and `supervisord-dev.conf`) and with cookiecutter there's no way to share template or macro code between files, but this global Python function is available to all templates.

And partially because the logic in this function (all the setting defaults and stuff) is really really hard to implement in Jinja2.

_shared/local_extensions.py
```diff
@@ -110,11 +113,74 @@ def include_exists(self, context, path):
...
+
+        if not settings.get("programs"):
+            if context["cookiecutter"]["_directory"] == "pyramid-app":
+                settings["programs"] = {
+                    "web": {
+                        "settings": {
+                            "command": f"gunicorn --bind :{context['cookiecutter']['port']} --workers 1 --reload --timeout 0 --paste conf/development.ini",
+                        }
+                    }
+                }
```
seanh, line 152: Newly-generated Pyramid app projects will get a single gunicorn process by default.

_shared/local_extensions.py
```diff
@@ -110,11 +113,74 @@ def include_exists(self, context, path):
...
+            else:
+                settings["programs"] = {
+                    "app": {
+                        "settings": {
+                            "command": "python3 -m {context['cookiecutter']['package_name']}",
+                        }
+                    }
+                }
```
seanh, line 160: Newly-generated non-web apps get a process that just run's the package's `__main__.py` file.

_shared/local_extensions.py
```diff
@@ -110,11 +113,74 @@ def include_exists(self, context, path):
...
+
+        for program in settings["programs"].values():
+            program.setdefault("dev", True)
+            program.setdefault("prod", True)
+            program.setdefault("settings", {})
+            program["settings"].setdefault("stdout_events_enabled", True)
+            program["settings"].setdefault("stderr_events_enabled", True)
+            program["settings"].setdefault("stopsignal", "KILL")
+            program["settings"].setdefault("stopasgroup", True)
```
seanh, line 169: This is trivial in Python but really hard to do in Jinja2. I'm not sure if it's possible to do this concisely and elegantly in Jinja2.

_shared/local_extensions.py
```diff
@@ -110,11 +113,84 @@ def include_exists(self, context, path):
...
+            assert context.name.endswith("supervisord-dev.conf")
+            test = lambda program_settings: program_settings["dev"]
+
+        settings["programs"] = {
+            program_name: program_settings
+            for program_name, program_settings in settings["programs"].items()
+            if program_settings[env]
+        }
+
+        return settings
```
seanh, line 193: This Supervisor-specific method is implemented in Python partially because it's called by two templates (`supervisord.conf` and `supervisord-dev.conf`) and with cookiecutter there's no way to share template or macro code between files, but this global Python function is available to all templates.

And partially because the logic in this function (all the setting defaults and stuff) is really really hard to implement in Jinja2.

_shared/local_extensions.py
```diff
@@ -110,11 +113,81 @@ def include_exists(self, context, path):
...
+                program["settings"].setdefault("stdout_logfile", "NONE")
+                program["settings"].setdefault("stderr_logfile", "NONE")
+
+        settings["programs"] = {
+            program_name: program_settings
+            for program_name, program_settings in settings["programs"].items()
+            if program_settings["dev" if dev else "prod"]
+        }
+
+        return settings
```
seanh, line 190: There's two reasons why this Supervisor-specific method has to be implemented in Python:

1. It's a total pain to write this logic (setting default values etc) in Jinja. Ends up really ugly
2. With cookiecutter there's no way to shared template code or macros between files. But this Python function is available to all templates

_shared/local_extensions.py
```diff
@@ -110,11 +113,78 @@ def include_exists(self, context, path):
...
+
+        Return the contents of the project's .cookiecutter/includes/{path} JSON
+        file deserialized into a Python object.
+
+        Return None if the project has no .cookiecutter/includes/{path} file.
+        """
+        return json.load(self._open(context, path))
+
+    @pass_context
+    def supervisor_settings(self, context):
```
seanh, line 134: There's two reasons why this Supervisor-specific method has to be implemented in Python:

1. It's a total pain to write this logic (setting default values etc) in Jinja. Ends up really ugly
2. With cookiecutter there's no way to shared template code or macros between files. But this Python function is available to all templates

_shared/project/conf/supervisord.conf
```diff
@@ -3,27 +3,37 @@ nodaemon=true
 environment=PYTHONUNBUFFERED="1"
 logfile=/dev/null
 logfile_maxbytes=0
-{% if cookiecutter.get("_directory") == "pyapp" %}
+{% if include_exists("conf/supervisord.conf.json") %}
```
seanh, line 6: This block of "duplicated" template code has the same structure and logic in `supervisord-dev.conf` and `supervisord.conf` but differs in some ways:

* The `supervisord.conf` template reads `supervisord.conf.json` whereas the `supervisord-dev.conf` template reads `supervisord-dev.conf.json`
* The set of default settings and values injected into each `[program:foo]` section differs between dev and prod
* The default lists of programs used if the project has no `supervisord.conf.json` or `supervisord-dev.conf.json` file are different

Unfortunately with cookiecutter there's no way to share template code or macros between two different template files in the `cookiecutters` repo. I can't say 100% for sure that there isn't some unintended way that we can hack around this but I don't think we should, especially not just for this one case.

_shared/local_extensions.py
```diff
@@ -154,4 +167,4 @@ def _readlines(self, context, path):
 
         includes = Path(target_dir) / ".cookiecutter/includes"
         full_path = Path(includes) / path
```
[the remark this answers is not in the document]
seanh: :+1: https://github.com/hypothesis/cookiecutters/pull/91/commits/0b3520982072a5705f3d9f03b222acb01363a1fc

_shared/project/conf/supervisord-dev.conf
```diff
@@ -1,27 +1,37 @@
...
+{% else %}
 
-{{ include("conf/supervisord-dev.conf") -}}
+[program:app]
+command=python3 -m {{ cookiecutter.package_name }}
+stdout_events_enabled=true
+stderr_events_enabled=true
+stopsignal=KILL
+stopasgroup=true
+{% endif %}
```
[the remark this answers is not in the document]
seanh: :+1: Moved this into a macro:

* https://github.com/hypothesis/cookiecutters/blob/f3e93dc640da1198d146dcdb89bc019bc64a56dd/_shared/project/conf/supervisord-dev.conf#L1-L27
* https://github.com/hypothesis/cookiecutters/blob/f3e93dc640da1198d146dcdb89bc019bc64a56dd/_shared/project/conf/supervisord.conf#L1-L29

It's a shame that I have to write it twice. But I think :crossed_fingers: this situation is unique to the twin `supervisord[-dev].conf` files: two files that're almost but not exactly the same. Hoping that problem won't come up again, in which case it's probably not worth trying to figure out a custom solution to the repetition (since we can't use normal Jinja2 includes or imports).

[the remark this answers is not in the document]
seanh: > Could you include a default `conf/supervisord-dev.conf.json` for the various project types as a templated out piece of data instead of having two mechanisms?

I don't think this would work, no. When creating a new project we don't have any way to control what order cookiecutter renders the files in, so we wouldn't be able to ensure that `.cookiecutter/includes/supervisord-dev.conf.json` gets rendered before `conf/supervisord-dev.conf` does. There _might_ be something we can do with a cookiecutter pre-gen hook here?? But that feels like quite a costly hack.

Generally I think templates need to be written to work in the absence of include files from the project for two cases:

1. When generating a new project none of the include files will exist yet
2. When updating an existing project any include files might be missing from the new project. For example an include file might have been deleted from the project. Or perhaps a new include got added to the templates since the last time this project was updated

I think any hacks that we try to get around this reality are gonna end up pretty messy

_shared/local_extensions.py
```diff
@@ -110,11 +112,22 @@ def include_exists(self, context, path):
     def include(self, context, path, indent=0):
         """Return the lines from the project's .cookiecutter/includes/{path} or an empty string."""
         try:
-            lines = self._readlines(context, path)
+            file_ = self._open(context, path)
```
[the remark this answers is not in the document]
seanh: > `handle` or `file_handle` is a good way to avoid this.

Hmm. I know it's common term but I'm not really sure what "handle" means though. I think semantically the file object *is* the file (it's an object representing the file) rather than a handle to the file. In the same way that a date object _is_ a date not a "date handle".

These seem to be called "file objects" in the Python docs so maybe `file_object` or `file_obj`? Or just follow [the example in the docs](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files) and use `f`?

I'll go with `file_obj`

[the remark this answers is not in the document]
seanh: > Should we be doing `with self._open(context, path) as handle` here to ensure the handle is closed?

Yep, I think we should be: https://github.com/hypothesis/cookiecutters/pull/91/commits/9a54ad5c9e2e3ee537c99a551fce02735944b8a6

_shared/local_extensions.py
```diff
@@ -143,8 +156,8 @@ def dedent(self, s):
         """Return `s` dedented."""
         return dedent(s)
 
-    def _readlines(self, context, path):
-        """Return the lines from `path` in the project's .cookiecutter/includes dir."""
+    def _open(self, context, path):
+        """Return the file at  `path` in the project's .cookiecutter/includes dir."""
         target_dir = context["cookiecutter"].get("__target_dir__")
 
         if not target_dir:
```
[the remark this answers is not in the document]
seanh: I think honestly `_open()` either returns the file or raises `FileNotFoundError` if the file doesn't exist. The exception type alone is probably clear enough?

_shared/local_extensions.py
```diff
@@ -143,8 +156,8 @@ def dedent(self, s):
         """Return `s` dedented."""
         return dedent(s)
 
-    def _readlines(self, context, path):
-        """Return the lines from `path` in the project's .cookiecutter/includes dir."""
+    def _open(self, context, path):
+        """Return the file at  `path` in the project's .cookiecutter/includes dir."""
```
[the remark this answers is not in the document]
seanh: :+1:

_shared/local_extensions.py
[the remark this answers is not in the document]
seanh: Eh, whoops, that didn't work...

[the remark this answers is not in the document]
seanh: Had to fix that up, I think this now works: https://github.com/hypothesis/cookiecutters/pull/91/commits/0264422ba0cc7c7c03c7240c1e51f8fa6db8e9b3

PR #93, opened 2022-11-18
POC: Allow customization of `bin/create-db` and call it in `dev`

jon-betts:
One possible solution to adding a Metabase use in Report to all the DBs when they start for:

 * https://github.com/hypothesis/report/issues/49

This PR:

 * Calls `bin/create-db` in dev too (with the default DB name)
 * Adds a customisation point in `create-db` to add more commands

We could avoid the customization point by going off template, but without `bin/create-db` being called in dev this wouldn't have the desired effect.

## What else might we do?

 * We could go off template for both `tox.ini` and/or `bin/create-db` in Report. But `tox.ini` is a very complicated bit of the templating system and wouldn't be nice to keep in step by hand
 * There might be things we can do in `dockercompose` to run some scripts as Postgres starts, but we'd need the tests / functests DBs to exist already. (See https://hub.docker.com/_/postgres/ "Initialisation scripts")
 * Something else?

## Questions

Does this cause any particular problems? It seems like it should gracefully fail when calling `CREATE DATABASE` in the first command.

### DIFF
--- a/_shared/project/bin/create-db
+++ b/_shared/project/bin/create-db
@@ -6,3 +6,7 @@
 #
 #     create-db <DB_NAME>
 make services args="exec postgres psql -U postgres -c 'CREATE DATABASE $1;'" > /dev/null 2>&1 || true
+
+{% if include_exists("create_db/tail") %}
+    {{- include("create_db/tail", indent=4) -}}
+{% endif %}
--- a/_shared/project/tox.ini
+++ b/_shared/project/tox.ini
@@ -74,7 +74,7 @@ depends =
 {% if cookiecutter.get("db") == "yes" or include_exists("tox/allowlist_externals") %}
 allowlist_externals =
 {% if cookiecutter.get("db") == "yes" %}
-    tests,functests: sh
+    dev,tests,functests: sh
 {% endif %}
 {% if include_exists("tox/allowlist_externals") %}
     {{- include("tox/allowlist_externals", indent=4) -}}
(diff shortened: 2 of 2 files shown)


seanh (changes requested):
It feels a bit odd to me for `make dev` to call `bin/create-db` since that script's job is to create the given DB if it doesn't already exist and `make dev` uses the default `postgres` DB which always exists. `make test` and `make functests` do need to call `create-db` because they use separate DBs that don't exist by default.

Also feels a bit odd for `create-db` to be customizable and to be used to add Postgres users: that script is just supposed to be about creating DBs not initializing them or creating users or data etc.

Would a better fit for this be the `make db` command (which currently initializes the DB schema and upgrades it to the latest version), or `make devdata` (which is what we generally use to insert necessary things into the dev DB: for example LMS needs a user, organization and authclient etc to exist in the h DB and these are upserted by h's `make devdata` command). I guess it depends on whether the Postgres user that needs to be added is considered to be part of the "database schema" (`make db`) or part of the "development data" (`make dev`). But those are the commands whose job is to initialize the dev DB into the correct working state so that this app will work and collaborate with the other apps in the dev env.

I think we're probably better off *not* initializing DBs (creating or updating the schema, upserting dev data, etc) automatically when `make dev` is run. For example sometimes you need to test what happens when an app is running using an older version of the DB schema, if `make dev` always upgraded the schema to the latest version that wouldn't work. Similarly you may need to test what happens if LMS's user etc don't exist in h's DB, but if `make dev` always inserted the dev data you wouldn't be able to test that. I think speed is also part of the reason why `make dev` doesn't call `make db` and `make devdata` automatically.

jon-betts:
Hi, thanks for the comment. I think this is slightly different from what's come before, because this is required to be in place in the tests and functests in order to run. So if you wipe a DB and it's not created as part of the natural run of doing `make test` etc then it won't work.

I wonder if this is just suggesting this isn't a part of the environmental setup at all? I mean how is this going to happen in live? We could add the users by hand, but that's more great.

Maybe this is an in-process even that the scripts have to do for themselves?

PR #95, opened 2022-11-22
Add GitHub Actions deployment workflow templates

seanh:
## Problem

All our deployed applications have a `deploy.yml` workflow, for example here's [Via's `deploy.yml`](https://github.com/hypothesis/via/blob/main/.github/workflows/deploy.yml). This file is exactly the same in every app (except for some simple variables) but is currently not generated by the cookiecutter. As a result it's labor-intensive and error-prone to maintain these files.

Every app also needs to contain a `redeploy.yml` workflow, for example here's [Via's `redeploy.yml`](https://github.com/hypothesis/via/blob/main/.github/workflows/redeploy.yml). `redeploy.yml` hasn't been added to any projects other than Via yet, but it needs to be added to them all. Same problem: this file is exactly the same in every app except for some variables and is labor-intensive and error-prone to maintain without the cookiecutter.

There's also a bunch of information about the app's QA and production environments that's duplicated between `deploy.yml` and `redeploy.yml`. Any differences between the environments referenced in the two workflows would lead to confusing misbehaviour.

## Solution

This PR adds `deploy.yml` and `redeploy.yml` templates to the `pyapp` and `pyramid-app` cookiecutters. Both templates read an `environments.json` file that must be provided by the project. Here's what the `environments.json` would look like for Via:

```json
{
  "QA": {
    "github_environment_url": "https://qa-via.hypothes.is/https://en.wikipedia.org/api/rest_v1/page/pdf/Comet_Kohoutek",
    "aws_region": "us-west-1",
    "elasticbeanstalk_application": "via",
    "elasticbeanstalk_environment": "qa"
  },
  "QA_Edu": {
    "github_environment_name": "QA (Edu)",
    "github_environment_url": "https://hypothesis.instructure.com/courses/125/assignments/878",
    "aws_region": "us-west-1",
    "elasticbeanstalk_application": "lms-via",
    "elasticbeanstalk_environment": "qa"
  },
  "Production": {
    "needs": ["QA"],
    "github_environment_url": "https://via.hypothes.is/https://en.wikipedia.org/api/rest_v1/page/pdf/Comet_Kohoutek",
    "aws_region": "us-west-1",
    "elasticbeanstalk_application": "via",
    "elasticbeanstalk_environment": "prod"
  },
  "Production_Edu": {
    "needs": ["QA_Edu"],
    "github_environment_name": "Production (Edu)",
    "github_environment_url": "https://hypothesis.instructure.com/courses/125/assignments/882",
    "aws_region": "us-west-1",
    "elasticbeanstalk_application": "lms-via",
    "elasticbeanstalk_environment": "prod"
  }
}
```

Even though we haven't applied the cookiecutter to most of our applications yet, my intention is to use the cookiecutter to generate just the `deploy.yml` and `redeploy.yml` files in each project (not committing any of the other files that the cookiecutter adds or modifies) rather than manually adding `redeploy.yml` to every project. For examples of how I intend to do this see https://github.com/hypothesis/via/pull/829 and https://github.com/hypothesis/h-periodic/pull/262.

## Testing

This PR shouldn't make any changes to projects that don't have an `environments.json` file, for example <https://github.com/hypothesis/cookiecutter-pyramid-app-test>:

```console
cd /tmp
git clone https://github.com/hypothesis/cookiecutter-pyramid-app-test
cd cookiecutter-pyramid-app-test
make template checkout=add-deployment-workflows
git diff
```

If you add a `.cookiecutter/includes/.github/workflows/environments.json` file like the one above to `cookiecutter-pyramid-app-test` then re-run `make template checkout=add-deployment-workflows` you should see that it adds correct `deploy.yml` and `redeploy.yml` files.

### DIFF
--- a/_shared/project/.github/workflows/deploy.yml
+++ b/_shared/project/.github/workflows/deploy.yml
@@ -1,0 +1,35 @@
+name: Deploy
+concurrency:
+  group: deploy
+  cancel-in-progress: true
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+jobs:
+  ci:
+    name: CI
+    uses: ./.github/workflows/ci.yml
+  docker_hub:
+    name: Docker Hub
+    needs: [ci]
+    uses: hypothesis/workflows/.github/workflows/dockerhub.yml@main
+    with:
+      Application: {% raw %}${{ github.event.repository.name }}{% endraw +%}
+    secrets: inherit
+  {% for job_id, environment in include_json(".github/workflows/environments.json", {}).items() %}
+  {{ job_id }}:
+    name: {{ environment.github_environment_name }}
+    needs: [{{ (["docker_hub"] + environment.get("needs", []))|join(", ") }}]
+    uses: hypothesis/workflows/.github/workflows/deploy.yml@main
+    with:
+      operation: deploy
+      github_environment_name: {{ environment.get("github_environment_name", job_id) }}
+      github_environment_url: {{ environment.github_environment_url }}
+      aws_region: {{ environment.aws_region }}
+      elasticbeanstalk_application: {{ environment.elasticbeanstalk_application }}
+      elasticbeanstalk_environment: {{ environment.elasticbeanstalk_environment }}
+      docker_tag: {% raw %}${{ needs.Docker_Hub.outputs.docker_tag }}{% endraw +%}
+    secrets: inherit
+  {% endfor %}
--- a/_shared/hooks/post_gen_project.py
+++ b/_shared/hooks/post_gen_project.py
@@ -38,6 +38,14 @@ def remove_conditional_files():
     paths_to_remove.extend(["tests/functional/cli_test.py"])
     {% endif %}
 
+    {% if not include_exists(".github/workflows/environments.json") %}
+    # The deployment workflows only make sense if environments.json exists.
+    paths_to_remove.extend([
+        ".github/workflows/deploy.yml",
+        ".github/workflows/redeploy.yml",
+    ])
+    {% endif %}
+
     for path in paths_to_remove:
         if os.path.exists(path):
             try:
--- a/_shared/local_extensions.py
+++ b/_shared/local_extensions.py
@@ -117,7 +117,7 @@ class LocalJinja2Extension(Extension):
             return ""
 
     @pass_context
-    def include_json(self, context, path):
+    def include_json(self, context, path, default=None):
         """Return the project's .cookiecutter/includes/{path} data or None.
 
         Return the contents of the project's .cookiecutter/includes/{path} JSON
@@ -125,8 +125,11 @@ class LocalJinja2Extension(Extension):
 
         Return None if the project has no .cookiecutter/includes/{path} file.
         """
-        with self._open(context, path) as file_obj:
-            return json.load(file_obj)
+        try:
+            with self._open(context, path) as file_obj:
+                return json.load(file_obj)
+        except FileNotFoundError:
+            return default
 
     def oldest(self, python_versions):
         """Return the oldest from `python_versions` by version number."""
--- a/_shared/project/.github/workflows/redeploy.yml
+++ b/_shared/project/.github/workflows/redeploy.yml
@@ -1,0 +1,29 @@
+name: Redeploy
+concurrency:
+  group: deploy
+  cancel-in-progress: true
+on:
+  workflow_dispatch:
+    {% if include_exists(".github/workflows/environments.json") %}
+    inputs:
+      {% for job_id, environment in include_json(".github/workflows/environments.json").items() %}
+      {{ job_id }}:
+        type: boolean
+        description: Redeploy {{ environment.get("github_environment_name", job_id) }}
+      {% endfor %}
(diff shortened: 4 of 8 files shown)


_shared/project/.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+    with:
+      Application: {% raw %}${{ github.event.repository.name }}{% endraw +%}
+    secrets: inherit
+  {% if include_exists(".github/workflows/environments.json") %}
+  {% for job_id, environment in include_json(".github/workflows/environments.json").items() %}
+  {{ job_id }}:
+    {% if "github_environment_name" in environment %}
+    name: {{ environment.github_environment_name }}
+    {% endif %}
+    needs: [{{ (["Docker_Hub"] + environment.get("needs", []))|join(", ") }}]
```
jon-betts, line 26:
```suggestion
    needs: [{{ (["Docker_Hub"] + environment.get("needs", [])) | join(", ") }}]
```

jon-betts (approved):
I've tried this out with the provided instructions and it works well. I think the JSON config works well. It's hard to see if it produces the "correct" output without having a product to compare it to. We can work all that out when it comes to integrating with a given product where the diff will show us any changes.

I think it was originally suggested that using JSON might be a solution for more complex files like `tox.ini` where you might have a large number of entry points as well. Now this is in the product it might be worth thinking if there are other places it could be applied.

Adding a default to `include_json` would mean you don't have to call `include_exists` immediately above it whenever it's used, which would cut down a bit on template noise.

_shared/project/.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+  group: deploy
+  cancel-in-progress: true
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+jobs:
+  CI:
+    uses: ./.github/workflows/ci.yml
```
jon-betts, line 12:
```suggestion
  ci:
    uses: ./.github/workflows/ci.yml
    name: Continuous Integration
```

The casing here is a bit strange. It's more obvious on following actions, but I think using a mixed camel case underscore makes it a bit more ambiguous how the file works. Are we linking by id or name? Especially here as we skip the name.

_shared/project/.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+  Docker_Hub:
+    name: Docker Hub
+    needs: CI
```
jon-betts, line 15:
```suggestion
  docker_hub:
    name: Docker Hub
    needs: [ci]
```

Using the brackets here isn't essential but would make this section look more like those generated from the JSON stuff below.

_shared/project/.github/workflows/deploy.yml
seanh: I'm gonna follow the style used in the Jinja2 docs and continue to not put spaces around `|`'s

_shared/project/.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+    uses: hypothesis/workflows/.github/workflows/dockerhub.yml@main
+    with:
+      Application: {% raw %}${{ github.event.repository.name }}{% endraw +%}
+    secrets: inherit
+  {% if include_exists(".github/workflows/environments.json") %}
+  {% for job_id, environment in include_json(".github/workflows/environments.json").items() %}
```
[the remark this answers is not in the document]
seanh: :+1: Done, this works.

Unfortunately I've got a feeling this won't work always work and in many cases `{% if include_exists("...") %}` may actually be required. I think we may only be getting away with it in this particular case because the entire contents of the `{% if %}` is just a `{% for %}`.

_shared/project/.github/workflows/deploy.yml
seanh: :+1: https://github.com/hypothesis/cookiecutters/pull/95/commits/24ed13d0f8d92911f5cd78c71b2eeb028f9a9905

This came about because if a GitHub Actions job doesn't have a `name` then the UI uses the job ID as the display name. So if you can have IDs like `CI`, `QA` and `Production` then you don't need `name`'s and the display names will still look nice (gotta have my capital letters). But it doesn't work for jobs like `Production (Canada)` because you can't have spaces and parens in job IDs. At least not easily, I didn't check whether it can be done with some horrible escaping, especially not because we sometimes need to use a job ID multiple times (when one job depends on another) and because our job IDs live in a JSON file that gets read in and rendered by Jinja2. So in cases like `Production (Canada)` I ended up with IDs like `Production_Canada` and then doing `name: "Production (Canada)"` for display. So you end up with some jobs that have a `name` and some that don't.

I think you're right that it's better to use `lowercase_with_underscores` IDs and always use `name`. It's less weird, more consistent, and makes the template code simpler.

_shared/project/.github/workflows/deploy.yml
seanh: :+1: https://github.com/hypothesis/cookiecutters/pull/95/commits/a0e1ea391ab6524864dba1e0d0404b5984cb46f0

PR #96, opened 2022-11-22
Add the `make sql` command to give handy access to the DB

jon-betts:
For:

 * https://github.com/hypothesis/report/issues/84


_shared/project/Makefile
```diff
@@ -44,6 +44,13 @@ shell: python
 	@pyenv exec tox -qe dev
 {% endif %}
 
+{% if cookiecutter.get("db") == "yes" %}
+.PHONY: sql
+$(call help,make sql,"Connect to the dev database with a psql shell")
+sql: python
+	@tox -qe dockercompose -- exec postgres psql --pset expanded=auto -U postgres
+{% endif %}
+
```
seanh, line 53: This is adding an unwanted blank line if you run `make template checkout=add-make-sql` in a project that doesn't have `"db"` (e.g. https://github.com/hypothesis/cookiecutter-pyramid-app-test). I think this should fix it:

```suggestion
{% if cookiecutter.get("db") == "yes" %}
.PHONY: sql
$(call help,make sql,"Connect to the dev database with a psql shell")
sql: python
	@tox -qe dockercompose -- exec postgres psql --pset expanded=auto -U postgres

{% endif %}
```

seanh (changes requested):

_shared/project/Makefile
jon-betts: Took me a couple of looks there to work that out. For what ever reason Github doesn't show the trailing space in the "from" side of the diff?

seanh (approved):

PR #97, opened 2022-12-01
Fix up services support in cookiecutters

seanh:
## Problem

The cookiecutter lacks services and DB support sufficient for any of our services-using or DB-using projects:

* The data-tasks package can't use the cookiecutter's `db` option to add a DB to that project because the `db` option only works with the `pyapp` and `pyramid-app` cookiecutters not `pypackage`.
* The cookiecutter can't be applied to projects like h, LMS, Checkmate, and others that have non-Postgres services because there's no way to include non-Postgres services in `docker-compose.yml` or `ci.yml`, the templates for these files provide no way to include non-Postgres services.

## Solution

This PR extends the cookiecutter's services and DB support enough to handle the services and DB needs of our various projects.

## Examples:

Draft PRs based on this branch that add services to various projects:

* https://github.com/hypothesis/cookiecutter-pyramid-app-test/pull/42
* https://github.com/hypothesis/cookiecutter-pyapp-test/pull/5
* https://github.com/hypothesis/cookiecutter-pypackage-test/pull/27
* https://github.com/hypothesis/data-tasks/pull/5

## Testing

Generate a Pyramid app with a DB:

```bash
cookiecutter gh:hypothesis/cookiecutters --checkout fix-services-support --directory pyramid-app --no-input db=yes
```

Test our various commands like `make services`, `make sql`, and `make sure`. Try adding `.cookiecutter/includes/docker-compose/postgres.yml`, `.cookiecutter/includes/docker-compose/services.yml`, `.cookiecutter/includes/docker-compose/tail.yml`, and `.cookiecutter/includes/.github/workflows/ci/services.yml` files and running `make template checkout=fix-services-support` and looking at the results.

## Notes

* A project can put `"db": "yes"` in `cookiecutter.json` and this will cause a `docker-compose.yml` file with a Postgres service to be generated, a `make services` command to be added to `Makefile`, a `services` section with a Postgres service to be added to `ci.yml`, etc. This works for both apps and packages, so packages like `data-tasks` can have services as well.
* `__db_port` in `cookiecutter.json` can be used to customise the Postgres port
* `__postgres_version` in `cookiecutter.json` can be used to customise the Postgres version
* `docker-compose.yml` can be customised using three raw text includes:
  1. `docker-compose/postgres.yml`, for adding extra lines to the bottom of the Postgres service definition. For example h and LMS stuff have network stuff here.
  2. `docker-compose/services.yml`, for adding extra lines to the bottom of the `services` section of `docker-compose.yml`. For example to add non-Postgres services that the cookiecutter doesn't know anything in particular about, like Elasticsearch and RabbitMQ.
  3. `docker-compose/tail.yml`, for adding extra lines to the very bottom of the `docker-compose.yml` file. For example h and LMS have some networking stuff here.
* Similarly the `services` section in `ci.yml` can be customised with a raw text `.github/workflows/ci/services.yml` include.

It's tempting to try to define all the services once in a `services.json` include and have that JSON data be used by both the `docker-compose.yml` and `ci.yml` templates. But I think this would be quite difficult to do. The `services` sections in `docker-compose.yml` and `ci.yml` can contain all kinds of arbitrary YAML data including resursively nested dicts and lists (with increasing indentation) and who knows what else that Docker Compose and GitHub Actions might support (YAML is a hell of a thing). I think we might have to write some fairly hardcore recursive Jinja2 code to render an arbitrary and potentially arbitrarily nested `services.json` file to YAML. And we'd have to write that Jinja2 code twice since with cookiecutter there's no way to share template code between the `docker-compose.yml` and `ci.yml` templates. We may also run into issues with JSON being unable to express things that can be expressed in the much more complicated YAML.

### DIFF
--- a/_shared/project/.github/workflows/ci.yml
+++ b/_shared/project/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       postgres:
         image: postgres:{{ cookiecutter.get("__postgres_version", "11.5-alpine") }}
         ports:
-        - {{ cookiecutter['__db_port'] }}:5432
+        - {{ cookiecutter['__postgres_port'] }}:5432
       {% endif %}
       {% if include_exists(".github/workflows/ci/services.yml") %}
         {{- include(".github/workflows/ci/services.yml", indent=6) -}}
@@ -28,7 +28,7 @@ jobs:
     {% endif %}
     {% if job in ["tests", "functests"] and cookiecutter.get("postgres") == "yes" %}
     env:
-      TEST_DATABASE_URL: postgresql://postgres@localhost:{{ cookiecutter['__db_port'] }}/{{ cookiecutter.package_name }}_test
+      TEST_DATABASE_URL: postgresql://postgres@localhost:{{ cookiecutter['__postgres_port'] }}/{{ cookiecutter.package_name }}_test
     {% endif %}
     {% if cookiecutter._directory == 'pypackage' and (job == 'tests' or job == 'functests') %}
     strategy:
@@ -61,7 +61,7 @@ jobs:
       {% endif %}
       {% if job in ["tests", "functests"] and cookiecutter.get("postgres") == "yes" %}
       - name: Create test database
-        run: psql -U postgres -h localhost -p {{ cookiecutter['__db_port'] }} -c 'CREATE DATABASE {{ cookiecutter.package_name }}_test'
+        run: psql -U postgres -h localhost -p {{ cookiecutter['__postgres_port'] }} -c 'CREATE DATABASE {{ cookiecutter.package_name }}_test'
       {% endif %}
       {% if cookiecutter.get("frontend") == "yes" %}
       - run: yarn install --frozen-lockfile
--- a/_shared/project/docker-compose.yml
+++ b/_shared/project/docker-compose.yml
@@ -3,7 +3,7 @@ services:
   postgres:
     image: postgres:{{ cookiecutter.get("__postgres_version", "11.5-alpine") }}
     ports:
-      - "127.0.0.1:{{ cookiecutter['__db_port'] }}:5432"
+      - "127.0.0.1:{{ cookiecutter['__postgres_port'] }}:5432"
     healthcheck:
       test: ["CMD", "pg_isready", "-U", "postgres"]
       interval: 1s
--- a/pyapp/cookiecutter.json
+++ b/pyapp/cookiecutter.json
@@ -14,7 +14,7 @@
     "devdata": ["no", "yes"],
     "postgres": ["no", "yes"],
     "docker": ["no", "yes"],
-    "__db_port": "{{ random_port_number() }}",
+    "__postgres_port": "{{ random_port_number() }}",
     "__docker_namespace": "{{ cookiecutter.github_owner }}",
     "__docker_network": "{{ cookiecutter.package_name }}_default",
     "__github_url": "https://github.com/{{ cookiecutter.github_owner }}/{{ cookiecutter.slug }}",
--- a/pyramid-app/cookiecutter.json
+++ b/pyramid-app/cookiecutter.json
@@ -17,7 +17,7 @@
     "__frontend_typechecking": "{{ cookiecutter.frontend }}",
     "postgres": ["no", "yes"],
     "docker": ["no", "yes"],
-    "__db_port": "{{ random_port_number() }}",
+    "__postgres_port": "{{ random_port_number() }}",
     "__docker_namespace": "{{ cookiecutter.github_owner }}",
     "__docker_network": "{{ cookiecutter.package_name }}_default",
     "__github_url": "https://github.com/{{ cookiecutter.github_owner }}/{{ cookiecutter.slug }}",
--- a/pypackage/cookiecutter.json
+++ b/pypackage/cookiecutter.json
@@ -12,7 +12,7 @@
     "postgres": ["no", "yes"],
     "create_github_repo": ["no", "yes"],
     "dependabot_pip_interval": ["monthly", "weekly", "daily"],
-    "__db_port": "{{ random_port_number() }}",
+    "__postgres_port": "{{ random_port_number() }}",
     "__entry_point": "{{ cookiecutter.slug }}",
     "__github_url": "https://github.com/{{ cookiecutter.github_owner }}/{{ cookiecutter.slug }}",
     "__pypi_url": "https://pypi.org/project/{{ cookiecutter.slug }}",


_shared/hooks/post_gen_project.py
```diff
@@ -27,8 +27,8 @@ def remove_conditional_files():
     paths_to_remove.extend(["bin/make_devdata"])
     {% endif %}
 
-    {% if cookiecutter.get("services") != "yes" %}
-    paths_to_remove.extend(["docker-compose.yml", "requirements/dockercompose.in"])
+    {% if not (cookiecutter.get("postgres") == "yes" or include_exists("docker-compose/services.yml")) %}
+    paths_to_remove.extend(["docker-compose.yml"])
```
seanh, line 31: This means that the cookiecutter will add the `docker-compose.yml` file to the project if _either_ `"postgres"` is `"yes"` in `cookiecutter.json` _or_ `.cookiecutter/includes/docker-compose/services.yml` exists.

_shared/hooks/post_gen_project.py
```diff
@@ -141,7 +141,6 @@ def main():
             "{{ cookiecutter.package_name }}/__init__.py",
             "{{ cookiecutter.package_name }}/app.py",
             "Dockerfile",
-            "docker-compose.yml",
```
seanh, line 144: This removes `docker-compose.yml` from `template_ignore_patterns` which means that `make template` will now update `docker-compose.yml`. (On `main` the cookiecutter will generate the initial `docker-compose.yml` file when creating a new project but `make services` will never update `docker-compose.yml`).

If a project doesn't _want_ `make template` to change `docker-compose.yml` it can just add `docker-compose.yml` to its `ignore` list in `cookiecutter.json`.

_shared/project/.github/workflows/ci.yml
```diff
@@ -14,14 +14,21 @@ jobs:
     needs: tests
     {% endif %}
     runs-on: ubuntu-latest
-    {% if cookiecutter.get("db") == "yes" %}
+    {% if cookiecutter.get("postgres") == "yes" or include_exists(".github/workflows/ci/services.yml") %}
```
seanh, line 17: Add the `services` section to `ci.yml` if _either_ `"postgres"` is `"yes"` in `cookiecutter.json` _or_ `.cookiecutter/includes/.github/workflows/ci/services.yml` exists.

_shared/project/Makefile
```diff
@@ -5,11 +5,11 @@ help = help::; @echo $$$$(tput bold)$(strip $(1)):$$$$(tput sgr0) $(strip $(2))
 $(call help,make help,print this help message)
 
 .PHONY: services
-{% if cookiecutter.get("services") == "yes" %}
+{% if cookiecutter.get("postgres") == "yes" or include_exists("docker-compose/services.yml") %}
```
seanh, line 8: Every project always gets a do-nothing `make services` command that is missing from `make help`. If _either_ `"postgres"` is `"yes"` in `cookiecutter.json` _or_ `.cookiecutter/includes/docker-compose/services.yml` then `make services` gets filled out with an actual `docker compose` command and a `make help` string.

_shared/project/Makefile
```diff
@@ -5,11 +5,11 @@ help = help::; @echo $$$$(tput bold)$(strip $(1)):$$$$(tput sgr0) $(strip $(2))
...
 $(call help,make services,start the services that the app needs)
 services: args?=up -d
 services: python
-	@tox -qe dockercompose -- $(args)
+	@docker compose $(args)
```
seanh, line 12: As of [Docker Desktop 4.12.0](https://docs.docker.com/desktop/release-notes/#4120) there's now a `docker compose` command that ships with Docker, so the `docker-compose` Python package is no longer needed and `make services` no longer needs to use tox.

_shared/project/tox.ini
```diff
@@ -104,7 +104,6 @@ commands =
     lint: pylint --rcfile=tests/pyproject.toml tests
     lint: pydocstyle {{ cookiecutter.package_name }} tests bin
     lint: pycodestyle {{ cookiecutter.package_name }} tests bin
-    dockercompose: docker-compose {posargs}
```
seanh, line 107: No more need for any `docker-compose` stuff at all in `tox.ini` now that `Makefile` just runs the `docker compose` command that ships with Docker Desktop instead. This is also the reason why `requirements/dockercompose.in` was removed.

pyapp/cookiecutter.json
```diff
@@ -12,8 +12,7 @@
     "dependabot_npm_interval": ["monthly", "weekly", "daily"],
     "dependabot_docker_interval": ["monthly", "weekly", "daily"],
     "devdata": ["no", "yes"],
-    "services": ["no", "yes"],
-    "db": ["no", "yes"],
+    "postgres": ["no", "yes"],
```
seanh, line 15: I removed the `services` option. It doesn't make sense for a project to say "Yes, I want services." The project actually has to say _which_ services it wants, otherwise the cookiecutter doesn't know which services to put in `docker-compose.yml` and `ci.yml`.

So instead of the `services` option there are now `docker-compose/services.yml` and `.github/workflows/ci/services.yml` includes. Rather than a boolean option, the existence of these include files is what triggers the `docker-compose.yml` file and the `services` section in `ci.yml` to be generated.

I also renamed the `"db"` option to `"postgres"`, seems like a better name.

_shared/project/README.md
```diff
@@ -72,14 +72,10 @@
...
-      Follow the [instructions on the Docker website](https://docs.docker.com/install/)
-      to install it.  
-      You **don't** need to install Docker Compose, the development environment
-      will install it automatically for you in tox.  
-      You **do** need to set up the `docker` command to work without `sudo`,
-      on Linux this means following Docker's [Post-installation steps for Linux](https://docs.docker.com/engine/install/linux-postinstall/).
+    {% if has_services() %}
+    * [Docker Desktop](https://www.docker.com/products/docker-desktop/).
+      On Ubuntu follow [Install on Ubuntu](https://docs.docker.com/desktop/install/ubuntu/).
+      On macOS follow [Install on Mac](https://docs.docker.com/desktop/install/mac-install/).
```
seanh, line 78: Docker Desktop seems to be the thing these days on both macOS and Linux, no longer Docker Engine. For example Docker Desktop is what the front page of docker.com now links to. It's also much easier to install and much nicer: it gives you a nice GUI Docker dashboard. This has already been [changed in the onboarding docs](https://github.com/hypothesis/onboarding/pull/65)

pyapp/cookiecutter.json
```diff
@@ -12,9 +12,9 @@
     "dependabot_npm_interval": ["monthly", "weekly", "daily"],
     "dependabot_docker_interval": ["monthly", "weekly", "daily"],
     "devdata": ["no", "yes"],
-    "services": ["no", "yes"],
```
seanh, line 15: I removed the `services` option. It doesn't make sense for a project to say "Yes, I want services." The project actually has to say _which_ services it wants, otherwise the cookiecutter doesn't know which services to put in `docker-compose.yml` and `ci.yml`.

So instead of the `services` option there are now `docker-compose/services.yml` and `.github/workflows/ci/services.yml` includes. Rather than a boolean option, the existence of these include files is what triggers the `docker-compose.yml` file and the `services` section in `ci.yml` to be generated.

_shared/local_extensions.py
```diff
@@ -214,6 +215,14 @@ def random_port_number(self):
 
         return self._random_port_numbers.pop()
 
+    @pass_context
+    def has_services(self, context):
+        """Return True if the project has services."""
+        cookiecutter = context["cookiecutter"]
+        return cookiecutter.get("db") == "yes" or self.include_exists(
+            context, "docker-compose/services.yml"
+        )
```
seanh, line 224: This boolean expression was going to get repeated again and again in the templates. We have no way to share code between template files except with the global template context functions in this `local_extensions.py` file. The repetition was particularly egregious and would have been a maintenance issue so I thought it was worth adding this boolean function to deduplicate it

jon-betts (approved):
Sorry for such a long comment, but I think it's worth trying our best to get this as right as is reasonable before it's in 25 projects and we can't face getting it back out.

I've no doubt this would work, but I'm a little unclear which specific problem we are trying to solve over what went before. The description says it's for a few things simultaneously and I'm not sure what the focus is.

A quick word about the YAML formatting: I think pasting in text is a totally fine way to go (it's what we do for most other things), but I think if we wanted a JSON file that rendered out as YAML we wouldn't bother writing a converter in jinja2, we'd use a serializer and splat out the output. This means the JSON would be 1:1 with the output YAML except... why bother? If we aren't going to map it, why not store the YAML as YAML as we do here and treat it as text? One answer is we could use the same data differently in another context (e.g. for ports), but then you are tied to the first format you chose. Anywho...

## Not having a docker-compose tox environment is cool

This is a nice benefit, and one less tox environment is good. This didn't work out of the box for me, so we should provide instructions when we release to prevent an avoidable annoyance for existing staff.

## Is Postgres a special case, or just another service?

One thing that jumped out at me is that we separate out Postgres for special treatment. Is this something that we could / should treat as a generic "service"? 

I think the answer is yes but we might want to extend the treatment to others. We can configure lots for you if we know Postgres is in the mix and there's a lot of standardisation about how we use it. I think it's appropriate to make special arrangements for Postgres. I'm not sure we have any other service which we use across more than one product consistently (NGINX?).

What I'm wondering is how we would onboard other services that did become standard to be more like the way we treat Postgres? I have some ideas for this:

  * Instead of `db=yes` and other files we could have `services=postgres,nginx,elasticsearch`
  * This way our test changes from `if db == 'yes'` to `if 'postgres' in services` (treated as a list not a string somehow)
  * Our check for whether we have services at all becomes `if services`
  * We could have `.cookiecutter/includes/docker-compose/service/<service_name>.yml` (and maybe `<service_name>-tail.yml`?)
  * We could then load them appropriate or provide defaults
  * This would keep the special treatment for Postgres, but make it available to others too
  * I think this is a nice split the difference between knowing there are services and being able to have as much custom behavior for each as we like
  
## I think we might benefit from embracing templating

It's not stated anywhere, but I think there is an attempt to make our templated files appear as if they are not templated / to minimally disrupt existing templates. It's only a guess, so I could be wrong.

I think we also might be attempting to ensure we have the minimal number of boilerplate files in a given project. So absent = off etc.

Again, not sure if these are aims or not, but if they are I think we could benefit from relaxing about both in the following ways:

 * Our templated output needs to be syntactically valid / shouldn't be horrible to read
 * I don't think it's required to ensure non-functional white-space (above and beyond function and basic readability) would match what we'd do by hand
 * I think it's good / acceptable to put comments in that indicate that these are templated files and show how the includes work
 * I think it would be good to include the full set of available include files (but blank / with comments) so you know they exist and where to find them
 * I think it would be good to put comments in them with their file name, so they show up in-situ in templated output as documentation / help
 
What this would mean:

 * Where it's annoying or complicated to write templates to match non-functional style requirements we can skip it
 * We can provide hints to let people know that a file is templated (right now there's no way to tell)
 * We can provide hints to show you how the available configuration entry-points (right now there's no way to know they are there)
 * This is especially true if the entry points change over time
 
## Putting these together will help with discoverabilty
 
At the moment, if you generated a project without Elasticsearch and wanted to add it, there's a steep wall to working out what to do. You might look in the cookie cutter config, but there's nothing there.  When you compare the two files from a project with or without services, there's no difference you can spot. You might look in the includes directory, but there's nothing there either.

The correct thing to do is `touch .cookiecutter/includes/docker-compose/services.yml` to turn it on, which is a big leap. You also need to independently know to go to `.cookiecutter/includes/.github/workflows/ci/services.yml` which does not operate as a switch.

Imagine this experience instead: When you look in your cookie cutter config file you see `services=`. If you compare this with `h` you see `services=postres,elasticsearch,rabbit` so you guess it's `services=elasticsearch`. You generate the template and see `docker-compose.yml` appear. When you look inside you see comments like:

```
services:
    # Loaded from `.cookiecutter/includes/docker-compose/elasticsearch.yml`
    elasticsearch:
      image: hypothesis/elasticsearch:latest
      ports:
        - 9202:9200
      env:
         discovery.type: single-node
```

Now you know where to go to tweak it if it's wrong etc. Ditto `.github/workflows/ci/services.yml`. The diff would guide you in what to do.

_shared/project/Makefile
seanh: Yes, although we only need to do that when we apply the cookiecutter to projects, not right away when we move this PR. I think most of the time is on macOS where they've been using Docker Desktop since forever so this shouldn't break them anyway

_shared/project/docker-compose.yml
```diff
@@ -1,8 +1,19 @@
...
+    {% if include_exists("docker-compose/postgres.yml") %}
+      {{- include("docker-compose/postgres.yml", indent=4) -}}
+    {% endif %}
+  {% endif %}
+  {% if include_exists("docker-compose/services.yml") %}
+    {{- include("docker-compose/services.yml", indent=2) -}}
+  {% endif %}
+{% if include_exists("docker-compose/tail.yml") %}
+  {{- include("docker-compose/tail.yml") -}}
+{% endif %}
```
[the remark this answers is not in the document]
seanh: > I'd recommend we we change include to check if something exists before do anything, it would remove the need for the ifs.

`include()` actually does [return an empty string if the file doesn't exist](https://github.com/hypothesis/cookiecutters/blob/0c1809aa3c650d3b30f125fdd649b46cd4c596dd/_shared/local_extensions.py#L162), although I should probably remove that feature of `include()` because it's never used anymore.

This can't remove the need for the `{% if %}`'s unfortunately. I think this is similar to our previous discussion about wishing to indent the bodies of Jinja2 blocks: it's just not how Jinja2 works and we're going to have to accept that rather than fight it.

Specifically what's going on here is that a Jinja2 `{% if %}` renders _nothing_ if the expression is false: not a blank line, not a newline character, not whitespace, nothing. Whereas a Jinja2 `{{ expression }}` such as `{{ include("foo") }}` always renders something no matter what the `include()` function returns.

Currently if none of these includes exist then the `{% if %}`'s output nothing at all: not even a blank line, newline character, or any whitespace. The `interval: 1s` line (at the end of the `postgres` service definition) will be the last line in the file. Here's what `docker-compose.yml` looks like when `"db"` is `"yes"` in `cookiecutter.json` but none of the three include files exists:

```yml
services:
  postgres:
    image: postgres:11.5-alpine
    ports:
      - "127.0.0.1:5437:5432"
    healthcheck:
      test: ["CMD", "pg_isready", "-U", "postgres"]
      interval: 1s
```

Here's what that would look like if we removed the `{% if %}`'s and allowed `include()` to return empty strings for include files that don't exist:

```yml
services:
  postgres:
    image: postgres:11.5-alpine
    ports:
      - "127.0.0.1:5437:5432"
    healthcheck:
      test: ["CMD", "pg_isready", "-U", "postgres"]
      interval: 1s
    
  

```

There are three empty lines at the bottom of the file, two of which also have whitespace characters in them.

Here's what it would render if `include()` returned `None` rather than `""` when the file doesn't exist:

```yml
services:
  postgres:
    image: postgres:11.5-alpine
    ports:
      - "127.0.0.1:5437:5432"
    healthcheck:
      test: ["CMD", "pg_isready", "-U", "postgres"]
      interval: 1s
    None
  None
None
```

It actually renders the text `None` into the output file.

You can attempt to use `{{-`'s and `-}}`'s to strip away the unwanted whitespace but that just creates other problems. For example when an include file _does_ exist then `{{-`'s and `-}}`'s will still strip away any whitespace before or after the contents of the include files and run those contents into the lines before and after. There's no way in Jinja2 to say "strip away the unwanted whitespace around this `{{ include() }}`, but only if `include()` didn't return anything".

Trust me, I have tried to do this, and you will drive yourself crazy trying to get it to render correctly for each of the possible combinations of nearby include files existing or not existing or `cookiecutter.json` options being off or on etc. You can't get it right in all cases at once. Nor can you always ignore the problems: a bit of unwanted whitespace _might_ not be meaningful (depending on the output file type) but you'll often get lines and words being run together in ways that will certainly break the output file. The reason for all the trouble is that you're misunderstanding how Jinja2 works and fighting against: once you start using `{% if %}`'s it all just works easily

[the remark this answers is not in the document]
seanh: > I've separately recommending that we always include something (e.g. a comment) in the file so you know it exists, so there would be some output and the if would also not be necessary.

We discussed this one in Slack but for the record I don't think this works either. I don't want to litter our output files with `"foo include goes here"` comments. Not all file format allow comments in all locations. And I also don't think it's possible for the cookiecutter to ensure that a default or placeholder include file (containing a comment) always exists for every supported include file: I think there'd be lots of problems with trying to implement that and I don't think it'll ultimately be possible to get it working.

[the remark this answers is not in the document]
seanh: > I think we should also be kind of committed to not caring about non-functional whitespace changes.

Mentioned above but it's not only meaningless whitespace that we'd end up with if we started trying to call `{{ include() }}` without `{% if %}`'s. In some cases you end up running lines together. In some output files whitespace is actually meaningful. And in some cases you would get away with it.

I think we should follow the approach of making the output files look nice, neat and readable in terms of formatting, whitespace, blank lines, etc, and eat the cost of the templates being a bit ugly. This is for a few reasons. One is that I just want the files in our dev environments to be nice and neat. Another is that it's not actually possible for the templates themselves to look nice, for a number of reasons:

* Template files are a mix of Jinja2 and YAML, INI, TOML, or whatever the output file format is going to be. There's no way to make this nice: the mixing together of the two completely different syntaxes is just horrible
* As previously discussed you can't indent the bodies of Jinja2 `{% if %}`'s, `{% else %}`s, `{% for %}`'s, macros, etc without breaking the indentation of the output file and indentation is meaningful in most of the cookiecutter's output file formats
* In some cases you won't be able to get rid of `{% if %}`'s from the templates either because this would cause more than just unwanted whitespace (e.g. running lines together) or because whitespace is actually meaningful in the output file

So we could say "we don't care about meaningless whitespace in the output files" and remove some of the `{% if %}`'s from the templates. But this still wouldn't make the templates nice at all: they'd still be a horrible mix of Jinja2 and the output format's syntax, Jinja2 blocks would still not be indented, and in some cases `{% if %}`'s would still be needed. So now we'd have the worst of both worlds: both our templates and our output files would be ugly.

The templates actually _can't_ be pretty but the output files can be.

pyapp/cookiecutter.json
seanh: That's actually how it works currently. There's no need to ever retain an empty file. Take a look at https://github.com/hypothesis/cookiecutter-pypackage-test/pull/27/ for example: that PR just changes `"db"` to `"yes"` in `cookiecutter.json`, does _not_ add any include files, and re-runs `make template`, and all the services stuff (`docker-compose.yml` file etc) gets added to the project

_shared/local_extensions.py
seanh: > Seems sensible. I do wonder if the "no way to share code" is something we should investigate changing. It seems like quite a speed bump for nice maintainable templates.

I've looked into this one previously. It's a limitation of cookiecutter and not something we can work around, at least not without the cure being worse than the disease. I think if we want to change it we should send a PR to cookiecutter and get them to merge it and release a new version. But it's probably not worth the effort: our templates aren't actually suffering that much from this.

seanh: I don't think it makes sense to have a `services=yes` configuration option because it's not sensible for a project to just say "I want services", the project has to actually say _which_ service(s) it wants: Postgres, Elasticsearch, etc. If all the project has said is "I want services" then the cookiecutter cannot generate a valid `docker-compose.yml` file (there's nothing to put in the `services` section), it cannot generate a valid `services` segment within the `ci.yml` file, etc.

So the way it works is that instead of `services=yes` we have options for specific services, currently this is only `db=yes`. So the project says `db=yes` and the cookiecutter is able to output a valid `docker-compose.yml` file with a `postgres` service in it, a follow `services` section within the `ci.yml` file with a `postgres` service in it, etc.

seanh:
> Sorry for such a long comment, but I think it's worth trying our best to get this as right as is reasonable before it's in 25 projects and we can't face getting it back out.

For what it's worth the changes in this PR will only effect projects that use services: just a few mostly test projects at first, several projects in the longer run. Currently there are draft PRs to add services to the three test projects (https://github.com/hypothesis/cookiecutter-pyramid-app-test/pull/42, https://github.com/hypothesis/cookiecutter-pyapp-test/pull/5 and https://github.com/hypothesis/cookiecutter-pypackage-test/pull/27) and to data-tasks (https://github.com/hypothesis/data-tasks/pull/5). That's all for now. As we roll the cookiecutter out to our applications, those applications that use services will make use of the feature in this PR. But other projects won't be affected.

seanh:
> I've no doubt this would work, but I'm a little unclear which specific problem we are trying to solve over what went before. The description says it's for a few things simultaneously and I'm not sure what the focus is.

It might be worth clarifying what's currently on `main`. I'll edit the PR description to make this clearer. `main` currently doesn't have working services or DB support, and what this PR does is add services and DB support that should be good enough to meet the services and DB needs of our projects (including our apps when the cookiecutter gets applied to them).

data-tasks can't use the cookiecutter's `db` option on `main` to add a DB to that project because the `db` option on `main` only works with the `pyapp` and `pyramid-app` cookiecutters not `pypackage`. On `main` the cookiecutter can't be applied to projects like h, LMS, Checkmate etc that have non-Postgres services because there's no way to include non-Postgres services in `docker-compose.yml` or `ci.yml`.

This PR makes it so that all those projects can use the cookiecutter and have the services they need

seanh:
> A quick word about the YAML formatting: I think pasting in text is a totally fine way to go (it's what we do for most other things), but I think if we wanted a JSON file that rendered out as YAML we wouldn't bother writing a converter in jinja2, we'd use a serializer and splat out the output. This means the JSON would be 1:1 with the output YAML except... why bother? If we aren't going to map it, why not store the YAML as YAML as we do here and treat it as text? One answer is we could use the same data differently in another context (e.g. for ports), but then you are tied to the first format you chose. Anywho...

I'm not sure what's being suggested here. How could we serialize a JSON file into YAML without converting JSON to YAML? You're saying we'd find an off-the-shelf JSON&rarr;YAML library and use that?

Unfortunately the cookiecutter can't generally use third-party libraries. The way that cookiecutters are used is that the user installs cookiecutter (e.g. `pipx install cookiecutter`), the user runs a command like `cookiecutter gh:hypothesis/cookiecutters --directory pyapp`, and then the various Python files in our cookiecutters (like `post_gen_project.py` and `local_extensions.py`) get run by cookiecutter in cookiecutter's Python environment (for example `~/.local/pipx/venvs/cookiecutter` if you installed cookiecutter with `pipx`). `cookiecutter` is the only Python package that is installed in this venv (well, plus whatever `cookiecutter` itself depends on which isn't much). Our cookiecutter is just a template not a Python package. There's no way for it to declare dependencies on Python packages from PyPI.

So we're limited to the standard library which I don't think contains any YAML parser or serializer.

I think there _are_ actually a couple of ways that we could get around this limitation but I don't it'd be worth it.

seanh:
> ### Is Postgres a special case, or just another service?
>
> One thing that jumped out at me is that we separate out Postgres for special treatment. Is this something that we could / should treat as a generic "service"?
>
> I think the answer is yes but we might want to extend the treatment to others...

Yes. We absolutely want to single-out Postgres for special treatment as there's much more to it than just the Docker Compose service itself. There's `make sql`, there are scripts for creating the test DBs that get called automatically, there's CI support, etc. It would be an absolute pain for each Postgres-using project is the cookiecutter didn't have special Postgres support.

Yes also to <q>we **might** want to extend the treatment to others</q> with emphasis on the _might_. I think we could add boolean `cookiecutter.json` options for other services that we have multiple projects using like `elasticsearch` (although actually only h uses that) or `rabbitmq` or `nginx`. On the other hand if these other services don't need as much boilerplate (outside of the `docker-compose.yml` file itself) then maybe it's not worth it. We'll see.

> What I'm wondering is how we would onboard other services that did become standard to be more like the way we treat Postgres? I have some ideas for this:
>
> * Instead of `db=yes` and other files we could have `services=postgres,nginx,elasticsearch`
> * This way our test changes from `if db == 'yes'` to `if 'postgres' in services` (treated as a list not a string somehow)

You can't put actual lists in `cookiecutter.json` unfortunately because cookiecutter interprets them as [choice variables](https://cookiecutter.readthedocs.io/en/stable/advanced/choice_variables.html) (so it'll make the user choose _one_ of `postgres`, `nginx` or `elasticsearch` to be the value of the `services` option). What you _can_ do though is have a `cookiecutter.json` option whose value is a CSV string. This is what we do with [the `python_versions` option](https://github.com/hypothesis/cookiecutters/blob/0c1809aa3c650d3b30f125fdd649b46cd4c596dd/pypackage/cookiecutter.json#L6). Then we have a [`python_versions()` function](https://github.com/hypothesis/cookiecutters/blob/0c1809aa3c650d3b30f125fdd649b46cd4c596dd/_shared/local_extensions.py#L139-L144) (available to templates globally) that parses the CSV and returns the Python versions as a list of strings (or in fact a list of custom `PythonVersion` objects in this case).

So we'd have to do the same thing for services and have a global `services()` function and it'd be `{% if "postgres" in services() %}`.

> * Our check for whether we have services at all becomes `if services`

This would actually have to be `{% if cookiecutter.get("services") %}` if you want to treat the raw services CSV string as a boolean (probably not a good idea because things like whitespace-only strings would be truthy) or `{% if services() %}` if you want use the list of strings from the custom `services()` function I mentioned above. This PR currently has `{% if has_services() %}`.

> * We could have `.cookiecutter/includes/docker-compose/service/<service_name>.yml` (and maybe `<service_name>-tail.yml`?)

Yep. That's what we have currently for Postgres (`.cookiecutter/includes/docker-compose/postgres.yml`) so I think we'd add the same for any other services that we added specific support for.

* * *

I think I'd prefer sticking with the current `db=yes` boolean option (although I should probably rename it to `postgres=yes`) and then adding more boolean options like `elasticsearch=yes` if we want to in future. For a few reasons:

* Currently a CSV `services="postgres"` option would be weird because it would ostensibly be a list of comma-separated service names but in fact `"postgres"` is the only possible valid value. This may not be a temporary situation: I think it's entirely likely that Postgres will be the only service that we ever add support for

* It's simpler, not requiring the CSV parsing or a custom `services()` function. Writing `{% if cookiecutter.get("db") == "yes" %}` all the time is quite awkward, we could wrap that in a global boolean function `db()` if we wanted the template code to be a bit nicer

* There's less potential for invalid values. The value of the `"db"` option is either `"yes"` (interpreted as truthy) or anything else is considered falsey. (If we want that to be a bit more forgiving about the exact string value then we should add the aforementioned `db()` function.) A CSV string on the other hand can be invalid CSV or it can be valid CSV but get some of the service names wrong: `elastic_search` instead of `elasticsearch`, or putting `redis` in there when we don't actually support Redis, etc. We might end up wanting to write validation code for it.

* I think boolean options work better with the `cookiecutter.json` file, discoverability, and cookiecutter's interactive interface.

  You can look in any project's `cookiecutter.json` file and see that there's an option called `"db"` that is set to `"no"`, and you can just change that to `"yes"` and then re-run `make template`. This is better than seeing a `"services"` option whose value is `""` and having no idea that the value should be a CSV string or what service names are valid to put in there.

  When you run cookiecutter to create a new project it presents you with the interactive choices `no` or `yes` for the `db` option:

  ```
  Select db:
  1 - no
  2 - yes
  Choose from 1, 2 [1]: 
  ```

  This is the same as what it does for several other boolean options that the cookiecutter has. It's not possible to choose an invalid value. Also if you use cookiecutter's command line interface (`cookiecutter gh:hypothesis/cookiecutters --directory pyramid-app --no-input db=yes`) then it validates for us that the value of `db` is either `yes` or `no`.

  This is better than it asking the user to type in a comma-separated list of service names and the user having no way of knowing that they should enter a CSV string or what the valid service names are. We could try to give them a clue by having a default value like `postgres,elasticsearch,rabbitmq,nginx` but that would force the user to delete all those if they don't want all those services and the default (for example if you just run `cookiecutter gh:hypothesis/cookiecutters --directory pyramid-app --no-input`) would be to turn on _all_ the services. With our current boolean value the default is always `no`: you get a minimal project by default and can opt-in to each optional feature.

seanh:
> ## Putting these together will help with discoverabilty
>
> At the moment, if you generated a project without Elasticsearch and wanted to add it, there's a steep wall to working out what to do. You might look in the cookie cutter config, but there's nothing there. When you compare the two files from a project with or without services, there's no difference you can spot. You might look in the includes directory, but there's nothing there either.

I think there's a misunderstanding here FWIW. You'll see `"db": "no"` in `cookiecutter.json` and just change that to `"yes"` and re-run `make services`. When creating a new project cookiecutter's interactive interface will ask you whether you want `db` to be `yes` or `no`. If we add a similar `elasticsearch` boolean in future it'll work the same way.

> The correct thing to do is `touch .cookiecutter/includes/docker-compose/services.yml` to turn it on, which is a big leap.

I think this is the misunderstanding. You don't need to have a `services.yml` include in order to have services (to have a `docker-compose.yml` file etc). You just need to set `"db"` to `"yes"`. But if you _do_ have a `services.yml` then that _will_ turn on the `docker-compose.yml` file even if `"db"` is not `"yes"`. What the two `services.yml` includes are for is: _adding services that the cookiecutter knows nothing about_. Which currently is any services other than Postgres. `services.yml` is just a way for the project to inject arbitrary text into the `docker-compose.yml` file. The cookiecutter has certain things that it can do for you (like Postgres) and then if you want to add some arbitrary `docker-compose.yml` text of your own (whether that's to add more services or anything else) you can use the includes to do so.

> Imagine this experience instead: When you look in your cookie cutter config file you see `services=`. If you compare this with `h` you see `services=postres,elasticsearch,rabbit` so you guess it's `services=elasticsearch`.

I think this is actually _less_ discoverable than seeing `"postgres": "no"`, `"elasticsearch": "no"` and `"rabbitmq": "no"` in your project's `cookiecutter.json` and just having to change them to `"yes"` (and also cookiecutter's interactive interface presenting you with the three boolean options when creating new projects).

pyramid-app/cookiecutter.json
```diff
@@ -15,9 +15,9 @@
     "devdata": ["no", "yes"],
     "frontend": ["no", "yes"],
     "__frontend_typechecking": "{{ cookiecutter.frontend }}",
-    "services": ["no", "yes"],
     "db": ["no", "yes"],
     "docker": ["no", "yes"],
+    "__db_port": "{{ random_port_number() }}",
```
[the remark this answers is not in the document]
seanh: We do yes. I think we might be able to do something in `post_gen_project.py` to not render `__db_port` into `cookiecutter.json` if `db` isn't on

_shared/project/.github/workflows/ci.yml
```diff
@@ -14,14 +14,21 @@ jobs:
     needs: tests
     {% endif %}
     runs-on: ubuntu-latest
-    {% if cookiecutter.get("db") == "yes" %}
+    {% if job in ["tests", "functests"] and has_services() %}
     services:
+      {% if cookiecutter.get("db") == "yes" %}
       postgres:
-        image: postgres:11.5-alpine
+        image: postgres:{{ cookiecutter.get("__postgres_version", "11.5-alpine") }}
```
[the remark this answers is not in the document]
seanh: Yep.

And the thing is you can't have optional questions. So if things like `__postgres_version` and `__postgres_port` didn't have the `__`'s then even if the user said "no" to Postgres (which is the default) it would still go on to ask you what Postgres version and port you wanted :( Hence the approach I've taken here is that `__postgres_version` and `__postgres_port` are private options (so the interactive UI never asks you for them) but they have default values (including that value being randomly generated in the case of the port number). A similar approach has been used for other dependent options

_shared/project/docker-compose.yml
```diff
@@ -1,8 +1,19 @@
...
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
+      interval: 1s
+    {% if include_exists("docker-compose/postgres.yml") %}
+      {{- include("docker-compose/postgres.yml", indent=4) -}}
+    {% endif %}
+  {% endif %}
+  {% if include_exists("docker-compose/services.yml") %}
+    {{- include("docker-compose/services.yml", indent=2) -}}
+  {% endif %}
```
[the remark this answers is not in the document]
seanh: I think they're at least potentially necessary:

* For each service definition that the cookiecutter supports in `docker-compose.yml` (currently only `postgres`) we're gonna want an include for adding arbitrary lines to the bottom of that service definition (the lines being indented so as to be part of the Postgres service definition) so that the project can customise the service definition. Not sure whether any of our projects actually _will_ do this for the Postgres service, but the potential need is obvious. If we add special knowledge about any other services in future (e.g. `elasticsearch` or whatever) then each of those will get its own include too
* You want an include for the bottom of the `services` section of the `docker-compose.yml` file (the included lines being indented so as to be part of the `services` section). This is how projects can add their own arbitrary service definitions that the cookiecutter knows nothing about (currently needed for any services other than Postgres)
* You want a final include for the very bottom of the `docker-compose.yml` file (not indented, not part of the `services` section). For example h and LMS will use this last include to add some networking stuff to `docker-compose.yml`.

seanh:
Ok, I think I've gone through all of the points raised here.

I'm gonna rename the `db` option to `postgres` as I think that's a better name and matches with the `__postgres_version` and `__postgres_port` options. Then I'll merge this.

I think a lot of the discussion is beyond the scope of this PR (e.g. talking about the general mechanism of how `include()` works etc). I've enabled GitHub Discussions on this repo so we can have a place to discuss things like this: https://github.com/hypothesis/cookiecutters/discussions

I think the main suggestion that's within the scope of this PR is to change the `"db": "yes"` boolean into a `"services": `"postgres"` list. But I think that suggestion may be based on a misunderstanding that it's necessary to have an include file to enable the services feature, rather than just setting `"db"` to `"yes"`. I think the boolean options are better, some notes in this comment: https://github.com/hypothesis/cookiecutters/pull/97#issuecomment-1351694995

Issue #39, opened 2022-07-19
Restrict the range of randomly generated port numbers


seanh:
For example it's currently possible for it to generate a too-low number that doesn't work

PR #98, opened 2022-12-05
Fix random port number generation

seanh:
Fixes https://github.com/hypothesis/cookiecutters/issues/39.

This isn't perfect: it requires us to maintain a hard-coded list of port numbers to avoid. I'm not sure there's much we can do about that though.

Testing:

```bash
cookiecutter gh:hypothesis/cookiecutters --checkout fix-random-port-number-generation --directory pyramid-app --no-input
```

Check the randomly generated port number in `cookiecutter.json`, `Makefile`, `supervisord.conf` and `supervisord-dev.conf`.

### DIFF
--- a/_shared/local_extensions.py
+++ b/_shared/local_extensions.py
@@ -1,9 +1,53 @@
 import json
+import random
 import textwrap
 from pathlib import Path
 from textwrap import dedent
 
 from jinja2.ext import Extension, pass_context
+
+
+UNAVAILABLE_PORTS = (
+    *range(1, 1025),  # Ports <1025 are for privileged processes.
+    8080,  # Common alternative HTTP port.
+    5000,  # h's web process.
+    5001,  # h's websocket process.
+    5432,  # h's Postgres service.
+    5672,  # h's RabbitMQ service.
+    15672,  # h's RabbitMQ service.
+    9200,  # h's Elasticsearch service.
+    3000,  # The Hypothesis client's development server.
+    3002,  # Alternative port for the client's development server.
+    3001,  # The client's package server.
+    8000,  # Bouncer's web process.
+    9082,  # Via's web process.
+    9083,  # Via's NGINX process.
+    3032,  # Via HTML's uWSGI process.
+    9085,  # Via HTML's NGINX process.
+    9086,  # Via HTML's NGINX process.
+    9101,  # Via Test App's web process.
+    8001,  # LMS's web process.
+    48001,  # LMS's web-https process.
+    5433,  # LMS's Postgres service.
+    5674,  # LMS's RabbitMQ service.
+    15674,  # LMS's RabbitMQ service.
+    9099,  # Checkmate's web process.
+    5434,  # Checkmate's Postgres service.
+    5673,  # Checkmate's RabbitMQ service.
+    15673,  # Checkmate's RabbitMQ service.
+    4000,  # Report's Metabase service.
+    5435,  # Report's metabase-postgres service.
+    5436,  # Report's Postgres service.
+    5437,  # cookiecutter-pypackage-test's Postgres service.
+    9202,  # cookiecutter-pyapp-test's Elasticsearch service.
+    9800,  # cookiecutter-pyramid-app-test's web process.
+    5438,  # cookiecutter-pyramid-app-test's Postgres service.
+    9201,  # cookiecutter-pyramid-app-test's Elasticsearch service.
+    5675,  # cookiecutter-pyramid-app-test's RabbitMQ service.
+    15675,  # cookiecutter-pyramid-app-test's RabbitMQ service.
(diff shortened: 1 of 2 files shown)


_shared/local_extensions.py
```diff
@@ -158,6 +160,57 @@ def dedent(self, s):
...
+        return random.choice(
+            [
+                i
+                for i in range(1025, 10000)  # Ports <=1024 are privileged on Linux.
+                if i
+                not in (
+                    # Commonly-used port numbers to avoid.
+                    8080,  # Common alternative HTTP port.
+                    # Port numbers already used by existing Hypothesis apps.
+                    5000,  # h's web process.
```
seanh, line 173: The code formatting is Black's, not mine :man_shrugging: Including the lack of empty lines between the different sections of the list.

_shared/local_extensions.py
```diff
@@ -158,6 +160,61 @@ def dedent(self, s):
...
+                        9201,  # cookiecutter-pyramid-app-test's Elasticsearch service.
+                        5675,  # cookiecutter-pyramid-app-test's RabbitMQ service.
+                        15675,  # cookiecutter-pyramid-app-test's RabbitMQ service.
+                        5050,  # Publisher Account Test Site's web process.
+                        9000,  # BioPub's web process.
+                    )
+                ]
+            )
+
+        return self._random_port_numbers.pop()
```
seanh, line 216: In the future a single run of the cookiecutter may generate more than one random port number, e.g. a web port and a Postgres port, so we don't want it to ever generate the same random number twice within a single run.

_shared/local_extensions.py
```diff
@@ -158,6 +160,60 @@ def dedent(self, s):
...
+        if not hasattr(self, "_random_port_numbers"):
+            self._random_port_numbers = [
+                i
+                for i in range(1025, 10000)  # Ports <=1024 are privileged on Linux.
+                if i
+                not in (
+                    # Commonly-used port numbers to avoid.
+                    8080,  # Common alternative HTTP port.
+                    # Port numbers already used by existing Hypothesis apps.
+                    5000,  # h's web process.
```
seanh, line 174: The code formatting is Black's, not mine :man_shrugging: Including the lack of empty lines between the different sections of the list.

jon-betts: Yeah I don't like this, as I often try to visually group things and then `black` smashes them all together.

You can space them out with comments, but it's still a bit ugly.

_shared/local_extensions.py
```diff
@@ -158,6 +160,60 @@ def dedent(self, s):
...
         return dedent(s)
 
+    def random_port_number(self):
+        """Return a randomly-generated port number."""
+        if not hasattr(self, "_random_port_numbers"):
+            self._random_port_numbers = [
+                i
+                for i in range(1025, 10000)  # Ports <=1024 are privileged on Linux.
+                if i
+                not in (
```
jon-betts, line 170:
```suggestion
                # Ports <= 1024 are privileged on Linux.
                port for i in range(1025, 10000) if port not in (
```

It might be good to separate the logic from the data here by having a "BAD_PORTS" or "TAKEN_PORTS" list or something similar?

jon-betts (approved):
I think this works, but something was poking my brain a bit about the fact it's randomized on each run, and we'd need to update the list all the time.

I think it would be good to pull the data of the list of ports out as a constant, to make the code clearer and make the bit you need to maintain easier.

### A random flight of fancy

I was wondering if we could tackle the reproducability with something like this:

```python
def hash_based_port(string):
    port_range = 10000 - 1025

    digest = hashlib.md5()
    digest.update(string.encode('utf-8'))
    return divmod(int(digest.hexdigest(), 16), port_range)[1] + 1025
```

This method will suggest the same port for the same string, but fairly "randomly" distributed throughout the range. This doesn't take into account the need to avoid existing ports, or generate more than one.

You can achieve that with a generator like this:

```python
def suggest_ports(name):
    port_index = 0

    while True:
        port = None
        while not port or port in bad_ports:
            port = hash_based_port(f"{name}-{port_index}")
            port_index += 1

        yield port
```

You would then use that something like this:

```python
port_suggester = suggest_ports("my-product")

print(next(port_suggester))
print(next(port_suggester))
print(next(port_suggester))
```

This doesn't really solve all the issues, but it will:

 * Generate a sequence of ports
 * The sequence is always the same for the same inputs
 * It will avoid items in the list

We might find that by being deterministic and distributed we will happen to get away without updating the list for a while until we start getting collisions?

Anyway, not important, but I thought it was a fun problem to think about.

_shared/local_extensions.py
seanh: :+1: Done: https://github.com/hypothesis/cookiecutters/pull/98/commits/1502746ca513954212ab10c39aa0145554873124

PR #103, opened 2022-12-19
Add `sqlalchemy` and `psycopg2` dependencies and `DATABASE_URL` and `TEST_DATABASE_URL` envvars

seanh:
See: https://github.com/hypothesis/data-tasks/pull/5#issuecomment-1357461195

In https://github.com/hypothesis/cookiecutters/pull/97 I added Postgres support to the cookiecutters with a `postgres` boolean option in `cookiecutter.json` but I only tested that `make services` and `make sql` work, I didn't test that the Python code can actually connect to the project's DB.

In fact in order to connect to the DB the Python code needs the `sqlalchemy` and `psycopg2` dependencies and the `DATABASE_URL` and `TEST_DATABASE_URL` envvars.

With this PR the cookiecutter still doesn't actually generate any Python code or test that actually connects to the DB and verifies that it actually works. Maybe it should but I'm trying not to overreach here, not sure each project is gonna want to set up its DB handle in the same way. Perhaps we can consider adding that in a separate PR.

### DIFF
--- a/_shared/project/.github/workflows/ci.yml
+++ b/_shared/project/.github/workflows/ci.yml
@@ -26,10 +26,6 @@ jobs:
         {{- include(".github/workflows/ci/services.yml", indent=6) -}}
       {% endif %}
     {% endif %}
-    {% if job in ["tests", "functests"] and cookiecutter.get("postgres") == "yes" %}
-    env:
-      TEST_DATABASE_URL: postgresql://postgres@localhost:{{ cookiecutter['__postgres_port'] }}/{{ cookiecutter.package_name }}_test
-    {% endif %}
     {% if cookiecutter._directory == 'pypackage' and (job == 'tests' or job == 'functests') %}
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


_shared/project/.github/workflows/ci.yml
```diff
@@ -26,10 +26,6 @@ jobs:
         {{- include(".github/workflows/ci/services.yml", indent=6) -}}
       {% endif %}
     {% endif %}
-    {% if job in ["tests", "functests"] and cookiecutter.get("postgres") == "yes" %}
-    env:
-      TEST_DATABASE_URL: postgresql://postgres@localhost:{{ cookiecutter['__postgres_port'] }}/{{ cookiecutter.package_name }}_test
-    {% endif %}
```
seanh, line 32: This is no longer needed here because it's in `tox.ini` (and `ci.yml` runs the tests via tox).

marcospri (approved):
Yeah, this should get everything lined up out of the box :+1:

Issue #55, opened 2022-08-16
Support installing packages directly from GitHub (no PyPI)


seanh:
I don't think we'd want to do this for things like libraries, tox plugins, etc, but for local development tools you can install the package directly from GitHub instead of from PyPI. This means the package doesn't have to be published to PyPI, doesn't have to have a name that's unique on PyPI, and isn't polluting PyPI with a package that may only be of interest to Hypothesis.

To install a package directly from GitHub with `pipx`:

```terminal
$ pipx install git+https://github.com/hypothesis/testpilot.git
```

Thanks to `setuptools-scm` the `--version` command works nicely and reports the git commit SHA (the part after the `g`):

```terminal
$ testpilot --version
0.2.8.dev11+gf92437b
```

Upgrading is the same as for PyPI packages, this will upgrade to the latest `main` commit:

```terminal
$ pipx upgrade testpilot
```

Uninstalling is the same: `pipx uninstall testpilot`.

To add support for this we'd need to:

- [x] Add `pypi` as a "boolean" (`["no", "yes"]`) option to the `pypackage` cookiecutter's `cookiecutter.json`
- [x] Only render the PyPI badge in the README if the `pypi` boolean is `"yes"`
- [x] Change the install command in the README if the `pypi` boolean is `"yes"` to `pipx install git+...` as above
- [x] Only render the docs about releasing new versions of the project in the README if the `pypi` boolean is `"yes"`
- [x] Only include the `pypi.yml` workflow file if the `pypi` boolean is `"yes"`

PR #104, opened 2022-12-19
Add support for non-PyPI Python packages

seanh:
Add a boolean `pypi` option to `cookiecutter.json` and only include the stuff around publishing the package to PyPI if `pypi` is `yes`. Otherwise support installing the package directly from GitHub.

This enables us to create things like development command line tools and just install them directly from the head of the `main` GitHub branch, no need to think of a unique name and release them to PyPI (or to have releases at all).

Fixes https://github.com/hypothesis/cookiecutters/issues/55.

Unfortunately the new `pypi` setting that this project adds defaults to `no`, which means we're going to have to add `"pypi": "yes"` to the `cookiecutter.json`'s of all our existing packages which is a bit of a pain. But all boolean settings in `cookiecutter.json` default to `no` and it'd be undesirable to have some that default to `yes`: when you generate a new project cookiecutter prompts you interactively for each setting and each time offers `no` as the default, it'd be a total snare if it then suggested `yes` as the default for one of the settings. I think `no` is also a good default for everything: it means you get a minimal project by default and can opt-in to optional features one-by-one.

### Testing

- [x] Projects that **do** have `pypi` set to `yes` will be unaffected by this PR. For example if you clone https://github.com/hypothesis/cookiecutter-pypackage-test, add `"pypi": "yes"` to `cookiecutter.json`, and run `make template checkout=support-non-pypi-packages`, `make template` should not make any changes.

- If you generate a new project with `pypi=no` (`cookiecutter gh:hypothesis/cookiecutters --directory pypackage --checkout support-non-pypi-packages --no-input console_script=yes pypi=no`, the `pypi=no` isn't really necessary because it's the default) then:

  - [x] There'll be no PyPI badge in `README.md`
  - [x] The install command in `README.md` will be a command for installing the project directly from git: `pipx install git+https://github.com/hypothesis/foo.git` 
  - [x] The docs about how to release a new version of the package will be missing from `README.md`
  - [x] The `.github/workflows/pypi.yml` file will be missing
  
  If you then change `pypi` to `yes` in `cookiecutter.json` and run `make template checkout=support-non-pypi-packages` then:

  - [x] A PyPI badge will be added to the top of `README.md`
  - [x] The install command in `README.md` will change to a PyPI one
  - [x] Docs about how to release a new version of the package will be added to `README.md`
  - [x] `.github/workflows/pypi.yml` will be added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ test sure: test-pypackage test-pyapp test-pyramid-app
 
 .PHONY: test-pypackage
 test-pypackage:
-	@bin/make_test pypackage console_script=yes db=yes devdata=yes services=yes
+	@bin/make_test pypackage console_script=yes db=yes devdata=yes pypi=yes services=yes
 
 .PHONY: test-pyapp
 test-pyapp:
--- a/_shared/hooks/post_gen_project.py
+++ b/_shared/hooks/post_gen_project.py
@@ -31,6 +31,10 @@ def remove_conditional_files():
     paths_to_remove.extend(["docker-compose.yml"])
     {% endif %}
 
+    {% if cookiecutter.get("pypi") != "yes" %}
+    paths_to_remove.extend([".github/workflows/pypi.yml"])
+    {% endif %}
+
     {% if cookiecutter.get("console_script") != "yes" %}
     paths_to_remove.extend(["src/{{ cookiecutter.package_name }}/__main__.py"])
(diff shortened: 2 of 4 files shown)


marcospri (approved):
Looks good :+1: 

Not sure if agree with the rationale for the default being "no" (other defaults are also no) but it's not a bad default in any case.

seanh:
Rebased

PR #105, opened 2023-01-10
Don't update the copyright year

seanh:
Don't update the copyright year in the `LICENSE` file to the current year. Otherwise every year the cookiecutter is going to want to update the `LICENSE` file in every project.

If we actually _want_ to automate updating copyright years one day we can do that: just write a script the reads `cookiecutter.json`, changes `__copyright_year`, writes it back, and runs `make template`. [Commando](https://github.com/hypothesis/commando) or another tool could be used to run such a script on every project and send PRs. But I don't think we need to update these.

Anyway, this PR just prevents `make template` from making this change by itself, which I think is clearly undesirable

### DIFF
--- a/_shared/project/LICENSE
+++ b/_shared/project/LICENSE
@@ -1,6 +1,6 @@
 BSD 2-Clause License
 
-Copyright (c) {% now 'utc', '%Y' %}, {{ cookiecutter.copyright_holder }}
+Copyright (c) {{ cookiecutter.__copyright_year }}, {{ cookiecutter.copyright_holder }}
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


jon-betts (approved):
I think this is the correct behavior. As I understand it we don't / shouldn't want to update the copyright date ever really.

This is because the date provides the starting point of copyright protection (+50 / 70 years). If you "update" the value, you strip yourself of historical protection. This means should someone else have copied the work in the mean time, and has a prior date to us, it will appear as if we stole the work rather than the other way around.

I believe you can include the start date and an update date in a copyright statement (c) 2001-2023 when the work has been altered. The correct time to do this would probably be on commit, but that would be horrible.

Long story short, I think this is much better.

PR #106, opened 2023-01-11
Fix `ci.yml` README badge

seanh:
The first badge at the top of all our project README's is broken (see https://github.com/hypothesis/gha-token/blob/main/README.md for example). This was broken by a https://shields.io/ change, see: https://github.com/badges/shields/issues/8671


jon-betts (approved):
This seems to match the format in the linked issue, but it'll be easiest to see in play once it's been played over a tempate.

PR #107, opened 2023-01-16
Don't run workflows unnecessarily

seanh:
Use `paths-ignore` to skip running the `ci.yml` and `deploy.yml` workflows for pull requests where the files modified make it unnecessary to run the workflow.  For example: don't run the tests on a PR that only modifies the documentation, or don't deploy a PR that only modifies development dependencies.

This should save us a lot of unnecessary workflow runs. For example we get a lot of Dependabot PRs for test and development dependencies, this will skip deploying any of those PRs.

This is as we've already done for h: https://github.com/hypothesis/h/pull/7755

GitHub's docs on this:

* https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths
* https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet


_shared/project/.github/workflows/ci.yml
```diff
@@ -1,6 +1,27 @@
...
+      - 'docs/*'
+      - 'requirements/*.in'
+      - 'requirements/dev.txt'
+      - '.docker.env'
+      - '**/.gitignore'
+      - '.python-version'
+      - 'Dockerfile'
+      - 'LICENSE'
+      - '*.md'
+      - 'docker-compose.yml'
```
seanh, line 24: Some of the paths that are listed are quite low-value in that they're files that we don't update very often so they won't save us many workflow runs. Nonetheless, since we can make the change just once here in the cookiecutter and have it be automatically applied to all cookiecutter-using apps (and we can update this list of ignores in the same semi-automated way) I thought it worth going to the effort to list all the cookiecutter-generated paths that can be ignored.

robertknight (approved):
I suspect these lists are going to be a bit prone to getting out of date, even with cookiecutter's help. That might not cause many problems though since the result will likely just be an unnecessary CI run or deploy.

PR #108, opened 2023-01-19
Fix up the `paths-ignore` in `ci.yml`

seanh:
Whoops, forget to remember when adding `paths-ignore` to the `ci.yml` template that this template is also used for our Python packages, not just for our apps.

Only ignore the paths that are relevant to the project. For example Python packages don't have a `deploy.yml` file so there's no need to put `deploy.yml` in the `paths-ignore` setting for packages.

### DIFF
--- a/_shared/project/.github/workflows/ci.yml
+++ b/_shared/project/.github/workflows/ci.yml
@@ -4,24 +4,40 @@ on:
     paths-ignore:
       - '.cookiecutter/*'
       - '.github/dependabot.yml'
+{% if include_exists(".github/workflows/environments.json") %}
       - '.github/workflows/deploy.yml'
       - '.github/workflows/redeploy.yml'
+{% endif %}
+{% if cookiecutter._directory in ['pyapp', 'pyramid-app'] %}
       - 'bin/logger'
+{% endif %}
       - 'bin/make_python'
       - 'bin/make_template'
+{% if cookiecutter._directory == 'pyramid-app' %}
       - 'conf/development.ini'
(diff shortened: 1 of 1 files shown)


_shared/project/.github/workflows/ci.yml
```diff
@@ -4,24 +4,40 @@ on:
     paths-ignore:
       - '.cookiecutter/*'
       - '.github/dependabot.yml'
+{% if include_exists(".github/workflows/environments.json") %}
       - '.github/workflows/deploy.yml'
       - '.github/workflows/redeploy.yml'
+{% endif %}
+{% if cookiecutter._directory in ['pyapp', 'pyramid-app'] %}
```
robertknight, line 11: I assume there isn't a public API that provides the same info as `_directory`?

robertknight (approved):

_shared/project/.github/workflows/ci.yml
seanh: The leading `_` here doesn't actually signify "private" as it usually does in Python. It's a cookiecutter thing: when you run the `cookiecutter` command to generate a new project it walks you interactively through choosing the values for all the cookiecutter's settings (What versions of Python do you want to support? Do you want to have Docker Compose services? Etc). Settings whose names begin with a leading underscore are settings that this interactive UI will skip (just using the default value), but projects can still set them either with command line arguments when running `cookiecutter` to create the project or by modifying the project's `cookiecutter.json` file and running `make template`

PR #110, opened 2023-03-08
Fix `update_repos.yml` when run on a schedule

seanh:
The `update_repos.yml` workflow is failing when run automatically by its
monthly schedule. For example see:

https://github.com/hypothesis/cookiecutters/actions/runs/4298577433

But the workflow is working when run manually:

https://github.com/hypothesis/cookiecutters/actions/runs/4365452102

The reason is that `inputs.repos` defaults to
`'__all_cookiecuttered_repos__'` when the workflow is run manually but
when the workflow is run automatically by its schedule the default
doesn't apply and `inputs.repos` is set to `''`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI
 on:
   push:
   workflow_dispatch:
+  schedule:
+  - cron: '0 1 * * *'
 jobs:
   Test_pypackage:
     name: Test pypackage
--- a/.github/workflows/update_repos.yml
+++ b/.github/workflows/update_repos.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Find repos
         id: find_repos
         run: |
-          if [ "${{ inputs.repos }}" = "__all_cookiecuttered_repos__" ]
+          if [ "${{ inputs.repos || '__all_cookiecuttered_repos__' }}" = "__all_cookiecuttered_repos__" ]
           then
             echo "REPOS=$(bin/find_repos | xargs)" >> $GITHUB_OUTPUT
           else
--- a/_shared/project/.github/workflows/ci.yml
+++ b/_shared/project/.github/workflows/ci.yml
@@ -40,10 +40,8 @@ on:
 {% endif %}
   workflow_dispatch:
   workflow_call:
-  {% if cookiecutter._directory == 'pypackage' %}
   schedule:
   - cron: '0 1 * * *'
-  {% endif %}
 jobs:
   {% for job in ['format', 'lint', 'tests', 'coverage', 'functests'] %}
   {{ job|capitalize }}:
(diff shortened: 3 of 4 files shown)


seanh:
I've tried running the workflow manually on this branch to make sure that running it manually still works with this version of the workflow, seems to be fine: https://github.com/hypothesis/cookiecutters/actions/runs/4365537099

We'll have to wait a month to find out whether the fix really works for the scheduled version.

.github/workflows/ci.yml
```diff
@@ -2,6 +2,8 @@ name: CI
 on:
   push:
   workflow_dispatch:
+  schedule:
+  - cron: '0 1 * * *'
```
seanh, line 6: Let's run the cookiecutter's own `ci.yml` nightly as we do for other projects, so we get notified if it breaks.

_shared/project/.github/workflows/ci.yml
```diff
@@ -40,10 +40,8 @@ on:
 {% endif %}
   workflow_dispatch:
   workflow_call:
-  {% if cookiecutter._directory == 'pypackage' %}
   schedule:
   - cron: '0 1 * * *'
-  {% endif %}
```
seanh, line 46: Let's run `ci.yml` on `main` nightly for all our projects, not just packages. It's less necessary for apps since they have pinned dependencies and receive Dependabot PRs but I think the nightly run is still useful to provide a history of CI runs on `main` so we can see if it's passing or not and whether it broke.

_shared/project/pyproject.toml
```diff
@@ -12,6 +12,8 @@ filterwarnings = [
     "error", # Fail the tests if there are any warnings.
     "ignore:^find_module\\(\\) is deprecated and slated for removal in Python 3.12; use find_spec\\(\\) instead$:DeprecationWarning:importlib",
     "ignore:^FileFinder.find_loader\\(\\) is deprecated and slated for removal in Python 3.12; use find_spec\\(\\) instead$:DeprecationWarning:importlib",
+    "ignore:^pkg_resources is deprecated as an API:DeprecationWarning:pkg_resources",
+    "ignore:^Deprecated call to .pkg_resources\\.declare_namespace\\('.*'\\).\\.:DeprecationWarning:pkg_resources",
```
seanh, line 16: These `pkg_resources` warnings were breaking the cookiecutter's own tests since those tests use the cookiecutter to generate a project and run the project's `make sure`. This change will also add these ignores to all our projects which should fix them--they're all crashing with this warning currently.

.github/workflows/update_repos.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Find repos
         id: find_repos
         run: |
-          if [ "${{ inputs.repos }}" = "__all_cookiecuttered_repos__" ]
+          if [ "${{ inputs.repos || '__all_cookiecuttered_repos__' }}" = "__all_cookiecuttered_repos__" ]
```
seanh, line 48: This is the actual bug fix. Just turn an empty string into `'__all_cookiecuttered_repos__'`

marcospri (approved):

PR #111, opened 2023-04-05
Expose port in `Dockerfile`

seanh:
This is necessary to get the `Dockerfile` working correctly in production. The port is also mentioned in the `-p` argument of the `docker run` command in `Makefile`: both seem to be necessary.


indigobravo (approved):
+1

PR #112, opened 2023-04-05
Add a `/_status` endpoint to Pyramid apps

seanh:
This is needed to deploy the app to production: Elastic Beanstalk needs a working healthcheck endpoint otherwise deploying fails

### DIFF
--- a/pyramid-app/{{ cookiecutter.slug }}/tests/functional/app_test.py
+++ b/pyramid-app/{{ cookiecutter.slug }}/tests/functional/app_test.py
@@ -1,11 +1,21 @@
+import pytest
 from webtest import TestApp
 
 from {{ cookiecutter.package_name }}.app import create_app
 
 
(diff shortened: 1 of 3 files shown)


pyramid-app/{{ cookiecutter.slug }}/tests/functional/app_test.py
```diff
@@ -1,11 +1,21 @@
...
 
 
-def test_it():
-    app = TestApp(create_app({}))
-
+def test_index(app):
     response = app.get("/")
 
-    assert response.json == {"Hello": "Pyramid!"}
+    assert response.text == "Hello world!"
```
seanh, line 10: Just for variety I changed the default root endpoint from JSON to text how that we have a JSON `/_status` endpoint

indigobravo (approved):
+1

PR #113, opened 2023-04-05
Rename `cookiecutter-pyramid-app-test`

seanh:
The test app has been renamed from cookiecutter-pyramid-app-test to the slightly shorter test-pyramid-app.


indigobravo (approved):
+1

PR #116, opened 2023-04-05
Fix typos in Supervisor JSON includes

seanh:
Oops


jon-betts (approved):
This seems to swap the entry-points we add to allow projects to customise these values.

PR #129, opened 2023-07-08
Fix the cookiecutter's tests

seanh:
Oops: the DB option is called `postgres` not `db`, and there's no `services` option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,12 +12,12 @@ test sure: test-pypackage test-pyapp test-pyramid-app
 
 .PHONY: test-pypackage
 test-pypackage:
-	@bin/make_test pypackage console_script=yes db=yes devdata=yes pypi=yes services=yes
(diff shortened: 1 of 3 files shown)


jon-betts (approved):
This works, but it worked in main too, but I suppose it wasn't testing what we thought it was. It seems we ignore invalid / unexpected arguments, so this was setting up the templated test environments incorrectly.

This also includes some extra fixes discovered while debugging some local issues on my machine.

seanh:
> This works, but it worked in main too, but I suppose it wasn't testing what we thought it was. It seems we ignore invalid / unexpected arguments, so this was setting up the templated test environments incorrectly.

I think you can use a `pre_gen_project.py` hook to validate options, created an issue for that: https://github.com/hypothesis/cookiecutters/issues/142

PR #130, opened 2023-07-08
Enable running `alembic` without `-c`

seanh:
Move Alembic's config file path into an envvar in `tox.ini`. This enables you to run commands like:

    tox -e dev --run-command 'alembic upgrade head'

without having to type out the `-c conf/alembic.ini` argument every time.

### DIFF
--- a/_shared/project/tox.ini
+++ b/_shared/project/tox.ini
@@ -30,6 +30,7 @@ setenv =
     dev: NEW_RELIC_APP_NAME = {env:NEW_RELIC_APP_NAME:{{ cookiecutter.slug }}}
     dev: NEW_RELIC_ENVIRONMENT = {env:NEW_RELIC_ENVIRONMENT:dev}
 {% if cookiecutter.get("postgres") == "yes" %}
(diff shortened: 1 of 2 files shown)


_shared/project/tox.ini
```diff
@@ -30,6 +30,7 @@ setenv =
     dev: NEW_RELIC_APP_NAME = {env:NEW_RELIC_APP_NAME:{{ cookiecutter.slug }}}
     dev: NEW_RELIC_ENVIRONMENT = {env:NEW_RELIC_ENVIRONMENT:dev}
 {% if cookiecutter.get("postgres") == "yes" %}
+    dev: ALEMBIC_CONFIG = {env:ALEMBIC_CONFIG:{{ cookiecutter.package_name }}/migrations/alembic.ini}
```
marcospri, line 33: Is this not in conf/ instead of migrations/?

https://github.com/hypothesis/cookiecutters/blob/962060d3f62145840ea299a9a95c3ce706d2413d/_shared/project/conf/alembic.ini

seanh: D'oh, yep, well spotted. Fixed

jon-betts (approved):

PR #132, opened 2023-07-08
Fix `alembic.ini`

seanh:
Fix a couple of bugs in the `alembic.ini` template:

1. `slug` (e.g. `test-pyramid-app`) is just wrong here, you want `package_name` (e.g. `test_pyramid_app`.

2. `database_url` doesn't work here: this is ready by alembic not by us and alembic looks for `sqlalchemy.url`.


marcospri (approved):
Yeah :+1:  this looks like the correct approach

In LMS we seem to try very hard to use "database_url" even an alembic config option if I'm reading this correctly:

https://github.com/hypothesis/lms/blob/ea49d6a7f228be88c03e7ae656322b3097a739f3/lms/migrations/env.py#L26

PR #133, opened 2023-07-08
Upgrade the cookiecutter's default Postgres version

seanh:
...to the latest version that's available to us from RDS.

### DIFF
--- a/_shared/project/docker-compose.yml
+++ b/_shared/project/docker-compose.yml
@@ -1,12 +1,14 @@
 services:
   {% if cookiecutter.get("postgres") == "yes" %}
   postgres:
-    image: postgres:{{ cookiecutter.get("__postgres_version", "11.5-alpine") }}
+    image: postgres:{{ cookiecutter.get("__postgres_version", "15.3-alpine") }}
     ports:
       - "127.0.0.1:{{ cookiecutter['__postgres_port'] }}:5432"
     healthcheck:
       test: ["CMD", "pg_isready", "-U", "postgres"]
       interval: 1s
+    environment:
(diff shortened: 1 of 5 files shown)


pyapp/cookiecutter.json
```diff
@@ -14,6 +14,7 @@
     "devdata": ["no", "yes"],
     "postgres": ["no", "yes"],
     "docker": ["no", "yes"],
+    "__postgres_version": "15.3-alpine",
```
seanh, line 17: Oops, the `__postgres_version` option wasn't actually working because it was missing from all three `cookiecutter.json`'s

_shared/project/docker-compose.yml
```diff
@@ -1,12 +1,14 @@
...
   postgres:
-    image: postgres:{{ cookiecutter.get("__postgres_version", "11.5-alpine") }}
+    image: postgres:{{ cookiecutter.get("__postgres_version", "15.3-alpine") }}
     ports:
       - "127.0.0.1:{{ cookiecutter['__postgres_port'] }}:5432"
     healthcheck:
       test: ["CMD", "pg_isready", "-U", "postgres"]
       interval: 1s
+    environment:
+      POSTGRES_HOST_AUTH_METHOD: trust
```
seanh, line 11: Postgres 15 refuses to start up without this envvar

marcospri: I was checking and we are disabling the password requirement with this.

I thought that maybe we'd need to update some of the database URLs but we already do not include a password there (and I reckon rely on the fact that the default was the same as the user, postgres/postgres?)

marcospri (approved):

PR #136, opened 2023-07-08
Add `pytest-factoryboy` to the cookiecutter

seanh:
h and LMS both have their own custom ways of integrating `factory_boy`
into `pytest` fixtures and they're both different from each other,
meanwhile our other projects don't have `factory_boy` integrated with
`pytest` at all.

There is an "official" `pytest` `factory_boy` implementation: the
`pytest-factoryboy` package from the `pytest-dev` GitHub repo:

https://github.com/pytest-dev/pytest-factoryboy

Let's just use this rather than inventing our own. We're already using
it in `gh-pr-upsert` (one of our command line tools).

### DIFF
--- a/_shared/project/requirements/functests.in
+++ b/_shared/project/requirements/functests.in
@@ -4,6 +4,7 @@ pip-sync-faster
 httpretty
 pytest
 factory-boy
+pytest-factoryboy
 h-matchers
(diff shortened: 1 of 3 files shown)


marcospri (approved):
I'd probably need a reminder to actually use this style in projects that use it but looks good.

PR #137, opened 2023-07-08
Set `factory_boy`'s random seed correctly

seanh:
This is copy-pasted from some of our apps that already do this, but it
was missing from the cookiecutter so most of our projects don't do this
correctly.

This would normally go in `conftest.py` but I don't want the
cookiecutter to own that file: that would be a huge pain because
projects need to do a lot of per-project stuff in `conftest.py`. So
instead I'm having the cookiecutter add a pytest plugin to contain the
fixture. (We haven't run into this issue before because the cookiecutter
has never provided any pytest fixtures before.)

### DIFF
--- a/_shared/project/tests/pytest_plugins/factory_boy.py
+++ b/_shared/project/tests/pytest_plugins/factory_boy.py
@@ -1,0 +1,10 @@
+import factory.random
+import pytest
+
+
+@pytest.fixture(scope="session", autouse=True)
+def factory_boy_random_seed():
+    # Set factory_boy's random seed so that it produces the same random values
+    # in each run of the tests.
+    # See: https://factoryboy.readthedocs.io/en/latest/index.html#reproducible-random-values
+    factory.random.reseed_random("{{ cookiecutter.github_owner }}/{{ cookiecutter.slug }}")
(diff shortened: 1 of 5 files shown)


_shared/project/tox.ini
```diff
@@ -131,7 +132,7 @@ commands =
     lint: pycodestyle src tests bin
 {% endif %}
     tests: coverage run -m pytest --failed-first --new-first --no-header --quiet {posargs:tests/unit/}
-    functests: pytest --failed-first --new-first --no-header --quiet {posargs:tests/functional/}
+    functests: python -m pytest --failed-first --new-first --no-header --quiet {posargs:tests/functional/}
```
seanh, line 135: The pytest plugin discovery doesn't work unless you do this 🤷‍♂️

_shared/project/tests/pytest_plugins/factory_boy.py
```diff
@@ -0,0 +1,10 @@
+import factory.random
+import pytest
+
+
+@pytest.fixture(scope="session", autouse=True)
+def factory_boy_random_seed():
```
marcospri, line 6: We could use the same mechanism to do the same for the random package:

https://github.com/pytest-dev/pytest/issues/667#issuecomment-112206152

marcospri (approved):

PR #139, opened 2023-07-08
Add DB, models, factories and transaction manager to `make shell`

seanh:
Add handy `tm`, `db`, `m` / `models` and `f` / `factories` globals to `make shell` (`pshell`)'s Python environment.

### DIFF
--- a/_shared/hooks/post_gen_project.py
+++ b/_shared/hooks/post_gen_project.py
@@ -166,7 +166,6 @@ def main():
             "package.json",
             "yarn.lock",
             ".docker.env",
-            "{{ cookiecutter.package_name }}/pshell.py",
             "tests/__init__.py",
             "tests/conftest.py",
             "tests/factories/__init__.py",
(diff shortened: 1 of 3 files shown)


_shared/hooks/post_gen_project.py
```diff
@@ -160,7 +160,6 @@ def main():
             "package.json",
             "yarn.lock",
             ".docker.env",
-            "{{ cookiecutter.package_name }}/pshell.py",
```
seanh, line 163: Remove `pshell.py` from the cookiecutter's `template_ignore_patterns`, meaning that the `make template` commands in projects will now update `pshell.py` from the latest version of the template. I think we do want the cookiecutter to own `pshell.py`: I think we probably want this file to be the same in every project. Projects can always add `pshell.py` to their own per-project ignore lists in `cookiecutter.py` if they want to do their own thing

_shared/project/pyproject.toml
```diff
@@ -59,6 +59,9 @@ omit = [
     {% if cookiecutter.get("_directory") == "pyapp" %}
     "*/{{ cookiecutter.package_name }}/app.py",
     {% endif %}
+    {% if cookiecutter.get("_directory") == "pyramid-app" %}
+    "{{ cookiecutter.package_name }}/pshell.py",
+    {% endif %}
```
seanh, line 64: Add `pshell.py` to the coverage ignore list. I don't think there's much point in unit testing this file: it's only used by `make shell` in dev.

PR #141, opened 2023-07-10
Wait for healthchecks in `make services`

robertknight:
This enables starting apps with `make services dev` without potentially running into errors from the app due to services not being ready when the web server tries to connect.

Services must define a healthcheck [1] in `docker-compose.yml` for this to work.

The `--wait` flag is missing from the Docker Compose docs, but see https://github.com/docker/compose/pull/8777.

[1] https://docs.docker.com/compose/compose-file/05-services/#healthcheck


seanh (approved):
This will cause PRs to be sent to all cookiecutter-using projects that have services, so we may need to add missing healthchecks to some of those services when those PRs come.

PR #151, opened 2023-09-01
Run CI when the python version changes

marcospri:
Is this actually used in CI? Or only the version in the docker file matters?

In any case I reckon this changes so rarely that it's not worth any misunderstandings.


seanh (approved):
Oh yeah, `.python-version` isn't actually used in CI currently, that's probably why it's ignore right now.

IIRC GitHub Actions does actually have support for reading the Python version from the `.python-version` file but I didn't use it because it makes installing Python slower. But in theory we might change to that one day and then we would want to run CI when this file changes.

Anyway, I'm happy with either keeping this here or removing it

PR #157, opened 2023-10-25
Fix the `--version` option

seanh:
The `--version` option as implemented doesn't work with required arguments. This is how you're supposed to do it.

Testing
-------

**CI**: CI in the cookiecutter's repo actually generates a new package with a CLI using the cookiecutter and runs the unit and functests that the cookiecutter generates for the new package, which include both unit and functests for `--version` so this is tested on CI. (It also runs the new package's `lint` and `checkformatting`.)

**Local:** Running `make sure` in the cookiecutter's repo will do the same tests locally as CI.

**Manually:**

```terminal
cookiecutter gh:hypothesis/cookiecutters --output-dir /tmp/testing --directory pypackage --checkout fix--version-option --no-input console_script=yes
cd /tmp/testing/my-python-package
tox -e dev --run-command 'my-python-package --version'
```

### DIFF
--- a/pypackage/{{ cookiecutter.slug }}/src/{{ cookiecutter.package_name }}/cli.py
+++ b/pypackage/{{ cookiecutter.slug }}/src/{{ cookiecutter.package_name }}/cli.py
@@ -4,10 +4,11 @@ from importlib.metadata import version
 
 def cli(_argv=None):  # pylint:disable=inconsistent-return-statements
     parser = ArgumentParser()
-    parser.add_argument("-v", "--version", action="store_true")
+    parser.add_argument(
+        "-v",
+        "--version",
+        action="version",
+        version=version("{{ cookiecutter.slug }}"),
+    )
 
     args = parser.parse_args(_argv)
-
-    if args.version:
-        print(version("{{ cookiecutter.slug }}"))
-        return 0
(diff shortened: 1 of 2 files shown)


robertknight:
Pardon a dumb question, but what is the process for testing a change like this in a downstream project?

seanh:
Apologies, just added testing instructions

robertknight (approved):
LGTM, assuming CI passes.

pypackage/{{ cookiecutter.slug }}/src/{{ cookiecutter.package_name }}/cli.py
```diff
@@ -2,12 +2,13 @@
 from importlib.metadata import version
 
 
-def cli(_argv=None):  # pylint:disable=inconsistent-return-statements
+def cli(argv=None):
```
seanh, line 5: I'm not sure why `argv` was named with a leading underscore here, that's not necesary

pypackage/{{ cookiecutter.slug }}/src/{{ cookiecutter.package_name }}/cli.py
```diff
@@ -2,12 +2,13 @@
...
-        print(version("{{ cookiecutter.slug }}"))
-        return 0
+    parser.add_argument(
+        "-v",
+        "--version",
+        action="version",
+        version=version("{{ cookiecutter.slug }}"),
+    )
+
+    _args = parser.parse_args(argv)
```
seanh, line 14: Having to name `_args` with a leading underscore here because otherwise pylint complains about an unused variable. I do want the `cli.py` generated by the cookiecutter to call `parse_args()` because I want it to tee everything up nicely for the user to come along and add the code that is actually going to use `_args`. But the cookiecutter itself doesn't have anything to do with `_args`. The alternative would have been a pylint disable comment.

PR #162, opened 2024-01-04
Fix up DB initialization

seanh:
Applies the DB initialization changes from https://github.com/hypothesis/h/pull/8414 to the cookiecutter.


.github/workflows/ci.yml
```diff
@@ -58,6 +58,18 @@ jobs:
...
+        image: postgres:15.3-alpine
+        ports:
+        - 5439:5432
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
seanh, line 72: This is the cookiecutter's own CI. This CI job generates a new project using the `pyramid-app` cookiecutter and runs the new project's `make sure` command. It now needs to have a DB running on CI because commands run by `make sure` will now run `init_db.py` which'll try to connect to the DB.

Makefile
```diff
@@ -20,4 +20,4 @@ test-pyapp:
 
 .PHONY: test-pyramid-app
 test-pyramid-app:
-	@bin/make_test pyramid-app devdata=yes docker=yes frontend=yes postgres=yes
+	@bin/make_test pyramid-app devdata=yes docker=yes frontend=yes postgres=yes __postgres_port=5439
```
seanh, line 23: Forcing the `__postgres_port` in the tests to be equal to the one used in `ci.yml` (otherwise it generates a random port when creating a new project)

_shared/project/pyproject.toml
```diff
@@ -60,6 +60,9 @@ omit = [
     {% if cookiecutter.get("_directory") == "pyapp" %}
     "*/{{ cookiecutter.package_name }}/app.py",
     {% endif %}
+    {% if cookiecutter.get("_directory") == "pyramid-app" and cookiecutter.get("postgres") == "yes" %}
+    "*/{{ cookiecutter.package_name }}/scripts/init_db.py",
+    {% endif %}
```
seanh, line 65: Super ugly diff rendering here. This just adds `init_db.py` to the coverage ignore list.

bin/make_test
```diff
@@ -13,6 +13,7 @@ tempdir=$(mktemp -d)
 slug="test-project"
 cookiecutter . --directory "$1" --output-dir "$tempdir" --no-input slug="$slug" "${@:2}"
 cd "$tempdir"/"$slug"
+if [ -z "${CI+x}" ]; then make services; fi
```
seanh, line 16: This is the cookiecutter's own tests. Same as on CI, the tests now need to run `make services` to start the DB. This runs `make services` only if the envvar `CI` is not set, because CI uses GitHub Actions's services thing, we don't need to start the services in Docker Compose on CI.

marcospri (approved):

PR #167, opened 2024-05-02
Post to Slack when cookiecutter workflows fail

seanh:
This is the same as what we do in other repos. The **Update repos** workflow had failed three months in a row and no one had noticed, so it seems worth adding this to Slack.

### DIFF
--- a/.github/workflows/slack.yml
+++ b/.github/workflows/slack.yml
@@ -1,0 +1,18 @@
+name: Slack
+on:
+  workflow_run:
+    workflows: [CI, Update repos]
+    types: [completed]
+    branches: [main]
+jobs:
+  on-failure:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/slack.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches: [main]
+jobs:
+  on-failure:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+      - name: Post to Slack
+        uses: slackapi/slack-github-action@v1.24.0
+        with:
+          channel-id: 'C4K6M7P5E'
```
marcospri, line 15: Is this #prod-general? It might become a bit too noisy but if we move it to some other channel we might ignore them instead. We'll have to find the balance.

marcospri (approved):

PR #169, opened 2024-05-16
Include basic mypy configuration for type checking

marcospri:
See: https://github.com/hypothesis/test-pyramid-app/pull/201

### DIFF
--- a/_shared/project/pyproject.toml
+++ b/_shared/project/pyproject.toml
@@ -179,9 +179,11 @@ warn_redundant_casts = true
 warn_unused_ignores = true
 
 disable_error_code = [
-  {% if include_exists("mypy/disables") %}
-    {{- include("mypy/disables", indent=4) -}}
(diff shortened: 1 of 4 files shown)


_shared/project/pyproject.toml
```diff
@@ -173,6 +173,25 @@ good-names = [
...
+[tool.mypy]
+warn_unused_configs = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+
+disable_error_code = [
+  {% if include_exists("mypy/disables") %}
+    {{- include("mypy/disables", indent=4) -}}
+  {% endif %}
+]
```
seanh, line 185: Could put the whole `disable_error_code = [ ... ]` block inside the `{% if %}`, so that if there's no include file you don't get an empty `disable_error_code = []`.

seanh (approved):

PR #170, opened 2024-05-17
Include pyapp/models and use the shared file in pyramid-app

marcospri:
Currently pyapp with postgres generates migrations but not models.

Add models to pyapp and use the common location in pyramid-app.


seanh (approved):
This will add `models/__init__.py` to some projects that might not want it (e.g. `report`), but we can add the file to that project's ignore list in `cookiecutter.json`.

PR #173, opened 2024-05-29
Fix `bin/find_repos` by hard-coding repos

seanh:
The GitHub API search query that `bin/find_repos` uses to find all
cookiecutter'ed repos has stopped returning the correct results. As a
result, the monthly `update_repos.yml` script (which sends automated
"Apply updates from cookiecutter" PRs to all our repos) stopped working
a few months ago.

It seems like something changed or broke on GitHub's end, but looking at
their code search docs
(https://docs.github.com/en/search-github/searching-on-github/searching-code)
our search query still looks like it should work to me. I haven't been
able to figure out how to fix it yet.

So for now just hard-code the list of repos into the script. This way at
least the `update_repos.yml` workflow will be able to continue sending
PRs to keep these repos up-to-date with the cookiecutter, even if it
means we have to manually maintain this hard-coded list.

### DIFF
--- a/bin/find_repos
+++ b/bin/find_repos
@@ -1,4 +1,37 @@
 #!/bin/bash
 # Return the list of all cookiecuttered repos in the hypothesis GitHub organization.
 
-gh api -X GET search/repositories --paginate -f 'q=cookiecutter.json in:readme org:hypothesis archived:false -repo:hypothesis/hdev' -q '.items | .[] | .full_name'
+# This GitHub API search query has stopped returning the correct results.
+#
+# Until we can figure out how to fix it lets have this script just print a
+# hard-coded list of repos so that update_repos.yml can still send cookiecutter
+# update PRs to our repos, even if it means we have to manually maintain this
+# hard-coded list.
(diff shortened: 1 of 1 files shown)


bin/find_repos
```diff
@@ -1,4 +1,37 @@
...
+echo hypothesis/pyramid-sanity
+echo hypothesis/report
+echo hypothesis/test-pyapp
+echo hypothesis/test-pypackage
+echo hypothesis/test-pyramid-app
+echo hypothesis/testpilot
+echo hypothesis/tox-envfile
+echo hypothesis/tox-faster
+echo hypothesis/tox-recreate
+echo hypothesis/via
```
seanh, line 36: I'm fairly sure I got all the current repos here. I wanted to get the list from the logs of the last successful run of `update_repos.yml` but that run was so long ago that GitHub Actions has already deleted the logs.

I have a local script that lists clones all the Hypothesis organization's repos into a local directory, and I constructed this list by grepping that local directory for `.cookiecutter/cookiecutter.json` files.

marcospri (approved):

PR #181, opened 2024-09-02
Pin GHA's upload-artifact to 3.1.3

marcospri:
As a temporary workaround for the changes newer versions (>3.2.0) around not uploading hidden files

See: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1725285976308269


marcospri:
This might not be needed after all.

PR #187, opened 2024-10-28
Add *.tsbuildinfo files (TypeScript build cache) to .gitignore

robertknight:
These files started appearing after updating to TypeScript 5.6. See https://github.com/hypothesis/h/pull/8992.

### DIFF
--- a/_shared/project/.gitignore
+++ b/_shared/project/.gitignore
@@ -11,6 +11,7 @@ supervisord.pid
 .DS_Store
 .devdata*
 .eslintcache
+*.tsbuildinfo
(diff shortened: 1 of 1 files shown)


seanh (approved):
We could potentially make this conditional and only add it to projects that have a frontend (`{% if cookiecutter.get("frontend") == "yes" %}`) but that hasn't been done for other frontend-related files in `.gitignore` and it's probably not worthwhile doing. One benefit would be creating fewer automated PRs after merging this one, as it won't add this to every single Python project. Nonetheless, I say just merge this as-is :+1:

PR #190, opened 2025-01-09
Ignore datetime utcnow deprecation

mtomilov:


_shared/project/pyproject.toml
```diff
@@ -16,6 +16,7 @@ filterwarnings = [
     "ignore:^pkg_resources is deprecated as an API:DeprecationWarning:pyramid",
     "ignore:^Deprecated call to .pkg_resources\\.declare_namespace\\('.*'\\).\\.:DeprecationWarning:pkg_resources",
     "ignore:^'cgi' is deprecated and slated for removal in Python 3\\.13$:DeprecationWarning:webob",
+    "ignore:^datetime\\.datetime\\.utcnow\\(\\) is deprecated and scheduled for removal in a future version\\.:DeprecationWarning",
```
mtomilov, line 19: This one due to `webob` which sadly hasn't been updated 2021 where even the current main [has the issue](https://github.com/Pylons/webob/blob/39d5af3c797e7b867f152c2e8c979de42d029403/src/webob/response.py#L1270). And we use it quite a bit [across the repos](https://github.com/search?q=org%3Ahypothesis+webob+path%3A*.txt&type=code).

marcospri (approved):

PR #196, opened 2025-01-10
Add an option to use Ruff (instead of Pylint, Black, isort, etc)

seanh:
Add a `"linter": "ruff"` option to `cookiecutter.json` that uses ruff instead of pylint, black, isort, pycodestyle and pydocstyle.

Testing:

- [x] This PR adds the `linter=ruff` to the cookiecutter's tests, and CI is passing. (This means that when new packages, apps and Pyramid apps are generated from the cookiecutter with the `linter=ruff` option, `make sure` is passing in the freshly-generated projects.)
- [x] Check out any Hypothesis project and run `make template checkout=ruff-linter-seanh-2` _without_ making any changes to `cookiecutter.json` and it should _not_ make any changes to the project. I've tested this with test-pypackage, test-pyapp, and test-pyramid-app.
- If you add `"linter": "ruff"` to `cookiecutter.json` and run `make template checkout=ruff-linter-seanh-2` it should add working Ruff integration to the project. For projects with requirements files you'll have to run `make requirements`. You may have to run `make fix` and `make noqa` to fix and suppress Ruff warnings. And you should grep for `# pylint` comments and remove them. Other than that it should just work. Here are working PRs to change test-pypackage, test-pyapp and test-pyramid-app to ruff:
  - [x] https://github.com/hypothesis/test-pypackage/pull/83
  - [x] https://github.com/hypothesis/test-pyapp/pull/190
  - [x] https://github.com/hypothesis/test-pyramid-app/pull/269
- [x] Also here's a PR that applies it to Via: https://github.com/hypothesis/via/pull/1509

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,12 +12,12 @@ test sure: test-pypackage test-pyapp test-pyramid-app
 
 .PHONY: test-pypackage
 test-pypackage:
-	@bin/make_test pypackage console_script=yes devdata=yes postgres=yes pypi=yes
+	@bin/make_test pypackage console_script=yes devdata=yes postgres=yes pypi=yes linter=ruff
 
 .PHONY: test-pyapp
 test-pyapp:
-	@bin/make_test pyapp devdata=yes docker=yes postgres=yes
+	@bin/make_test pyapp devdata=yes docker=yes postgres=yes linter=ruff
 
 .PHONY: test-pyramid-app
 test-pyramid-app:
-	@bin/make_test pyramid-app devdata=yes docker=yes frontend=yes postgres=yes __postgres_port=5439
+	@bin/make_test pyramid-app devdata=yes docker=yes frontend=yes postgres=yes __postgres_port=5439 linter=ruff
--- a/_shared/hooks/post_gen_project.py
+++ b/_shared/hooks/post_gen_project.py
@@ -57,6 +57,10 @@ def remove_conditional_files():
     paths_to_remove.extend([".github/workflows/pypi.yml"])
     {% endif %}
 
+    {% if cookiecutter.get("linter") == "ruff" %}
+    paths_to_remove.extend(["tests/pyproject.toml", "setup.cfg"])
+    {% endif %}
+
     {% if cookiecutter.get("console_script") != "yes" %}
     paths_to_remove.extend(["src/{{ cookiecutter.package_name }}/__main__.py"])
     paths_to_remove.extend(["src/{{ cookiecutter.package_name }}/cli.py"])
@@ -102,6 +106,10 @@ def remove_project_files(target_dir):
         ".cookiecutter/includes/setuptools/install_requires",
         ".cookiecutter/includes/setuptools/console_scripts",
         ".cookiecutter/includes/setuptools/entry_points",
+        {% if cookiecutter.get("linter") == "ruff" %}"tests/pyproject.toml",{% endif +%}
+        {% if cookiecutter.get("linter") == "ruff" %}"setup.cfg",{% endif +%}
+        {% if cookiecutter.get("linter") == "ruff" %}".cookiecutter/includes/setup.cfg",{% endif +%}
+        {% if cookiecutter.get("linter") == "ruff" %}".cookiecutter/includes/pycodestyle",{% endif +%}
     ]
 
     for path in paths_to_remove:
--- a/_shared/project/Makefile
+++ b/_shared/project/Makefile
@@ -68,6 +68,18 @@ $(call help,make lint,"lint the code and print any warnings")
 lint: python
 	@pyenv exec tox -qe lint
 
+{% if cookiecutter.get("linter") == "ruff" %}
+.PHONY: fix
+$(call help,make fix,"apply fixes to resolve lint violations")
+fix: python
+	@pyenv exec tox -qe lint -- ruff check --fix-only {{ cookiecutter.package_name}} tests bin
+
+.PHONY: noqa
+$(call help,make noqa,"add noqa comments to suppress lint violations")
+noqa: python
+	@pyenv exec tox -qe lint -- ruff check --add-noqa {{ cookiecutter.package_name}} tests bin
+
+{% endif %}
 .PHONY: typecheck
 $(call help,make typecheck,"type check the code and print any warnings")
 typecheck: python
--- a/_shared/project/pyproject.toml
+++ b/_shared/project/pyproject.toml
@@ -59,6 +59,112 @@ filterwarnings = [
     {% endif %}
 ]
 
+{% if cookiecutter.get("linter") == "ruff" %}
+[tool.ruff]
+target-version = "py{{ python_versions()|oldest|pyformat(PyFormats.MAJOR_MINOR_FMT) }}"
+{% if include_exists("ruff/top_level/tail") %}
+{{ include("ruff/top_level/tail") -}}
+{% endif %}
+
+[tool.ruff.lint]
+select = ["ALL"]
+ignore = [
+{% if include_exists("ruff/lint/ignore") %}
+{{ include("ruff/lint/ignore", indent=4) -}}
+{% else %}
+    "UP",    # pyupgrade
+    "YTT",   # flake8-2020 (checks for misuse of sys.version or sys.version_info
+    "ANN",   # flake8-annotations (checks for absence of type annotations on functions)
+    "ASYNC", # flake8-async (checks for asyncio-related problems)
+    "S",     # flake8-bandit (checks for security issues)
+    "FBT",   # flake8-boolean-trap (checks for the "boolean trap" anti-pattern)
+    "B",     # flake8-bugbear (checks for bugs and design problems)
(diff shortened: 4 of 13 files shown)


Makefile
```diff
@@ -12,12 +12,12 @@ test sure: test-pypackage test-pyapp test-pyramid-app
 
 .PHONY: test-pypackage
 test-pypackage:
-	@bin/make_test pypackage console_script=yes devdata=yes postgres=yes pypi=yes
+	@bin/make_test pypackage console_script=yes devdata=yes postgres=yes pypi=yes linter=ruff
```
seanh, line 15: Use the Ruff option in the cookiecutter's tests

_shared/hooks/post_gen_project.py
```diff
@@ -45,6 +45,9 @@ def remove_conditional_files():
 
     {% if cookiecutter.get("linter") == "ruff" %}
     paths_to_remove.extend(["tests/pyproject.toml"])
+    {% if cookiecutter._directory in ['pyapp', 'pyramid-app'] %}
+    paths_to_remove.extend(["setup.cfg"])
```
seanh, line 49: Now that we're no longer using pycodestyle the `setup.cfg` file is no longer needed. We do still need `setup.cfg` in packages because it [contains a bunch of package metadata](https://github.com/hypothesis/test-pypackage/blob/c65d559fbe4cf915830adfd4a314a6494e0564a1/setup.cfg#L1-L29), though I wonder whether that could be moved into `pyproject.toml` nowadays (separate PR).

_shared/project/pyproject.toml
```diff
@@ -24,11 +24,13 @@ filterwarnings = [
 
 {% if cookiecutter.get("linter") == "ruff" %}
 [tool.ruff]
-target-version = "py311"
+target-version = "py{{ python_versions()|oldest|pyformat(PyFormats.MAJOR_MINOR_FMT) }}"
```
seanh, line 27: Don't hard-code the Python version. Notice that this sets Ruff's target version to the _oldest_ Python version that the project supports, which is the right thing to do (we don't want `make lint` telling us to use Python features that we can't use yet).

_shared/project/pyproject.toml
```diff
@@ -24,11 +24,13 @@ filterwarnings = [
...
 line-length = 88
+{% if include_exists("ruff/exclude") %}
 exclude = [
-    "tests/bdd/steps/_compiled_feature_steps.py",
+{{ include("ruff/exclude", indent=4) -}}
```
seanh, line 31: Use an include instead of hard-coding this

_shared/project/tox.ini
```diff
@@ -66,12 +66,16 @@ deps =
     pip-sync-faster
 {% elif cookiecutter._directory == 'pypackage' %}
     dev: ipython
+{% if cookiecutter.get("linter") == "ruff" %}
+    format,checkformatting,lint: ruff
+{% else %}
```
seanh, line 71: Packages don't use requirements files, instead the dependencies for formatting, linting, testing, etc are listed here in `tox.ini` (and not pinned)

_shared/project/tox.ini
```diff
@@ -135,6 +139,13 @@ commands =
     {% endif %}
 {% else %}
     dev: {posargs:ipython --classic --no-banner --no-confirm-exit}
+    {% if cookiecutter.get("linter") == "ruff" %}
+    format: ruff check --select I --fix src tests bin
+    format: ruff format src tests bin
+    checkformatting: ruff check --select I src tests bin
+    checkformatting: ruff format --check src tests bin
+    lint: ruff check --preview -q src tests bin
+    {% else %}
```
seanh, line 148: The commands are all duplicated twice in the `tox.ini` template: once for apps and once for packages. Here I'm adding Ruff support to the commands section for packages. (This could probably be refactored to remove the duplication.)

pyramid-app/{{ cookiecutter.slug }}/{{ cookiecutter.package_name }}/scripts/init_db.py
```diff
@@ -114,10 +113,7 @@ def main():
         stamped = is_stamped(engine)
 
     if args.create:
-        {% if cookiecutter.get("linter") == "pylint" %}
-        # pylint:disable=possibly-used-before-assignment
-        {% endif %}
-        if stamped:
+        if stamped:{% if cookiecutter.get("linter") == "pylint" %}  # pylint:disable=possibly-used-before-assignment{% endif +%}
```
seanh, line 116: Again just avoiding making an unnecessary change to non-Ruff projects

_shared/hooks/post_gen_project.py
```diff
@@ -102,6 +106,10 @@ def remove_project_files(target_dir):
         ".cookiecutter/includes/setuptools/install_requires",
         ".cookiecutter/includes/setuptools/console_scripts",
         ".cookiecutter/includes/setuptools/entry_points",
+        {% if cookiecutter.get("linter") == "ruff" %}"tests/pyproject.toml",{% endif +%}
+        {% if cookiecutter.get("linter") == "ruff" %}"setup.cfg",{% endif +%}
+        {% if cookiecutter.get("linter") == "ruff" %}".cookiecutter/includes/setup.cfg",{% endif +%}
+        {% if cookiecutter.get("linter") == "ruff" %}".cookiecutter/includes/pycodestyle",{% endif +%}
```
seanh, line 112: Automatically delete these files from projects

_shared/project/Makefile
```diff
@@ -68,6 +68,18 @@ $(call help,make lint,"lint the code and print any warnings")
...
+{% if cookiecutter.get("linter") == "ruff" %}
+.PHONY: fix
+$(call help,make fix,"apply fixes to resolve lint violations")
+fix: python
+	@pyenv exec tox -qe lint -- ruff check --fix-only {{ cookiecutter.package_name}} tests bin
+
+.PHONY: noqa
+$(call help,make noqa,"add noqa comments to suppress lint violations")
+noqa: python
+	@pyenv exec tox -qe lint -- ruff check --add-noqa {{ cookiecutter.package_name}} tests bin
```
seanh, line 80: These two are honestly pretty handy especially when migrating projects to Ruff. You can just run `make format fix noqa lint`.

_shared/project/pyproject.toml
```diff
@@ -59,6 +59,104 @@ filterwarnings = [
...
+    # > The summary line may be on the same line as the opening quotes or on
+    # > the next line.
+    # >
+    # > https://peps.python.org/pep-0257/#multi-line-docstrings
+    "D212", "D213",
+    "E501", # line-too-long (we use the code formatter so we don't need the linter to check line lengths for us).
+    "PLR2004", # "Magic value used in comparison", this mostly triggers false-positives related to HTTP status codes.
+    "PLR6301", # Method could be a function/classmethod/static method (doesn't use self)
+    "RET501", # Do not explicitly return None if it's the only possible return value.
+    "RET504", # Unnecessary assignment before return statement.
```
seanh, line 131: I went for selecting ALL rules and then explicitly ignoring some. The idea is that we'll remove rules from this ignore list in time. This is a large ignore list for now to make it easier to migrate projects to Ruff, but I think we should aim to get this down to a pretty small list.

_shared/project/pyproject.toml
```diff
@@ -59,6 +59,104 @@ filterwarnings = [
...
+{% if include_exists("ruff/lint/ignore") %}
```
marcospri, line 132: Shouldn't this have the ignore and ignore_tail patter like other similar entry points?

I'm thinking about ignoring a separate set of rules on other project, not just adding more ignores.

seanh: Yeah. I was thinking we might not have any projects that want to completely override the ignore list. But probably I should add that ability to the template just in case, before committing it. Will do...

seanh: Done. So we have:

* `includes/ruff/top_level/tail`: extends the default top-level Ruff config
* `includes/ruff/lint/ignore`: replaces the default ignore list
* `includes/ruff/lint/ignore/tail`: extends the default ignore list
* `includes/ruff/lint/per_file_ignores`: replaces the default per-file ignores
* `includes/ruff/lint/per_file_ignores/tail`: extends the default per-file ignores
* `includes/ruff/tail`: extends the very bottom of the Ruff section of `pyproject.toml`. The idea is you'd use this to add any entirely new `[tool.ruff.*]` blocks that you want and they'll be added to the config file following the other `[tool.ruff.*]` blocks.

There's no include for _replacing_ the default top-level config. But that config only contains the `target-version` anyway. We can always add `includes/ruff/top_level` in future.

The rest of the includes in `pyproject.toml` and throughout the cookiecutter are not so well thought-out or consistently named, unfortunately. We may want to come up with a better and consistent pattern for all of them. But that's for another day.

I think there may also be a way to reformat the `pyproject.toml` template using Jinja2 macros or functions that'd make the template code a lot easier to read and to maintain.

PR #201, opened 2025-01-17
Move setuptools project metadata into `pyproject.toml`

seanh:
Move setuptools project metadata from `setup.cfg` into `pyproject.toml`.

This won't allow us to remove the `setup.cfg` from our projects yet as
it still contains pycodestyle config, but once we've replaced
pycodestyle with ruff we'll be able to remove the `setup.cfg` file
entirely.

For docs on configuring this setuptools project metadata in
`pyproject.toml` see:

https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/


_shared/hooks/post_gen_project.py
```diff
@@ -66,6 +80,40 @@ def remove_conditional_files():
...
+
+    This feature should be using sparingly because it's not generally safe: it
+    could remove files that contain uncommitted changes.
+
+    """
+    paths_to_remove = [
+        ".cookiecutter/setuptools/install_requires",
+        ".cookiecutter/setuptools/console_scripts",
+        ".cookiecutter/setuptools/entry_points",
+    ]
```
seanh, line 105: Old includes that've now been deleted from the `setup.cfg` template. This will automatically remove these includes from any projects that were using them. Those projects will now need to change to the new includes from the `pyproject.toml` template.

marcospri (approved):

_shared/hooks/post_gen_project.py
```diff
@@ -66,6 +80,40 @@ def remove_conditional_files():
...
+
+    for path in paths_to_remove:
+        path_ = target_dir / path
+
+        if os.path.exists(path_):
```
mtomilov, line 110: Maybe it's worth using `pathlib.Path` methods directly for all the operations?

mtomilov (approved):

PR #210, opened 2025-01-23
Add entry point for the top level configuration

marcospri:
mypy/top_level/tail will append configuration options in the main mypy section of the pyproject.toml file.

Applied it on LMS with the right include: https://github.com/hypothesis/lms/pull/6967


seanh (approved):
:+1: I gave it a quick test using the three test projects (testing `make template` with and without the new include), seems to work as desired

PR #214, opened 2025-01-28
Enable many more Ruff linter rules

seanh:
...and switch from an opt-in to an opt-out strategy where we enable ALL
rules and then disable certain specific rules or sets of rules. This
means that when new versions of Ruff come out new rules will be enabled
by default.

Slack discussion: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1737635626518379

Draft PRs to apply this to apps:

* https://github.com/hypothesis/h/pull/9283
* https://github.com/hypothesis/lms/pull/6975
* https://github.com/hypothesis/via/pull/1513

### DIFF
--- a/_shared/project/pyproject.toml
+++ b/_shared/project/pyproject.toml
@@ -72,49 +72,9 @@ ignore = [
 {% if include_exists("ruff/lint/ignore") %}
 {{ include("ruff/lint/ignore", indent=4) -}}
 {% else %}
-    "UP",    # pyupgrade
-    "YTT",   # flake8-2020 (checks for misuse of sys.version or sys.version_info
     "ANN",   # flake8-annotations (checks for absence of type annotations on functions)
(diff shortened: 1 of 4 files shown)


_shared/project/pyproject.toml
```diff
@@ -153,10 +113,20 @@ ignore = [
     "PLR0913", # Too many arguments. Tests often have lots of arguments.
     "PLR0917", # Too many positional arguments. Tests often have lots of arguments.
     "PLR0904", # Too many public methods. Test classes often have lots of test methods.
+    "S101", # Use of `assert` detected.
```
seanh, line 116: We want to allow `assert`'s in tests, obviously.

marcospri (approved):
We might have to tweak the list as we are more familiar with them all but this looks like a pretty good start.

seanh:
Rebased

PR #216, opened 2025-01-31
Fix `make db`

seanh:
This was broken by https://github.com/hypothesis/cookiecutters/pull/215:
it turns out `make db` also relies on the `PYTHONPATH` hack, not just
`make shell`.

### DIFF
--- a/_shared/project/Makefile
+++ b/_shared/project/Makefile
@@ -18,7 +18,7 @@ $(call help,make db,initialize the DB and upgrade it to the latest migration)
 db: args?=upgrade head
 db: python
 	@tox -qe dev --run-command 'python3 -m {{ cookiecutter.package_name }}.scripts.init_db --create --stamp'
-	@tox -qe dev --run-command 'alembic $(args)'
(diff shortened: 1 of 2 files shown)


mtomilov (approved):

marcospri:
Running alembic directly like:

```tox -e dev --run-command 'alembic history'```

runs into the same issue, I can run it like:


```PYTHONPATH=. tox -e dev --run-command 'alembic history'```


Could we include PYTHONPATH on the default passenv for `dev`?

mtomilov:
Maybe in dev / db we could pass all env vars (or almost all env vars) from the outside? Not sure if `tox` supports "allow * and disable some" approach though.

seanh:
> Maybe in dev / db we could pass all env vars (or almost all env vars) from the outside? Not sure if `tox` supports "allow * and disable some" approach though.

Tempting. Reasons for listing envvars explicitly in `passenv` are partly that it prevents the user's envvars from messing things up, and partly because it provides documentation of all the envvars that the app can possibly use (in dev, anyway). But it is tempting to do try doing `passenv = *` and just never worry about it again.

PR #217, opened 2025-03-05
Git ignore VSCode files by default

marcospri:
I'm sorry to be that guy, but it's just a matter of time before I commit this otherwise.


mtomilov (approved):
Hehe makes total sense
Was surprised not to see `pycharm` folder in there but turns our it was ignored by my global `gitignore` file.
Might want to include it here myself at some point.

PR #218, opened 2025-03-13
Fix make devdata in LMS

marcospri:
While not every project's `make devdata` will need this LMS does

See 3294792 for more context

### DIFF
--- a/_shared/project/Makefile
+++ b/_shared/project/Makefile
@@ -28,7 +28,7 @@ devdata: python
 {% if cookiecutter.get("_directory") == "pyramid-app" and cookiecutter.get("postgres") == "yes" %}
 	@tox -qe dev --run-command 'python3 -m {{ cookiecutter.package_name }}.scripts.init_db --create --stamp'
 {% endif %}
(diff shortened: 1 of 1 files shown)


robertknight:
[Additional context in Slack](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1741859276899199)

robertknight (approved):
This is fine, although I wonder if there is a more general way of solving this in tox @seanh?

seanh:
> This looks fine, although I wonder if there is a more general way of solving this in tox @seanh?

I think the proper fix is that our applications should be installable Python packages in `src/` dirs and tox should install them into the virtualenv, which it doesn't currently. We do this for our _packages_ but not for our applications, see the `skip_install` setting in `tox.ini`. 

There are other reasons why people recommend doing this: it prevents a class of problems where something works locally but does not work in production because some files that're present locally do not get copied into the production Docker container. If dev env commands work by installing an app package into a venv, and production works by installing the same package into a venv in the Docker container, then they're the same.

And it would also mean that `import <app>` would work for any Python code running in said venv.

For now I think we should merge this hack-around :+1:

